PR #3, opened 2021-05-05
docs: improve some formulations in the themes documentation

nico-bachner:
I fixed some typos and reformulated some sentences in the themes docs.

Feel free to pick and choose among the larger changes — some are a matter of taste.


antfu:
Nice catch, thanks a lot! I have reverted some as they are on purpose.

PR #20, opened 2021-05-06
fix(theme): border-color

random-yang:
**Before:**
image
---
**After:**
image
---

I know there must be some better fix. But I am not very familiar with `Windi CSS` and other tool chain in this proj.


antfu:
Nice, and your solution is almost perfect 👍

However, I already noticed it and landed a fix https://github.com/slidevjs/slidev/commit/0bf192e19679521a3c272d77c4cce22bb4b2f5fd and it will be shipped with next release.

Sorry as this is duplicated, thanks for the contributions!

random-yang:
That's great! You are so efficient, bro!

PR #40, opened 2021-05-09
fix(client): nav control focus (Triggered with next page/animation at the same time)

FyisFe:
Hi Anthony,

I noticed that you have fixed this issue. However, the same problem also occurs if my mouse does not leave the navigation bar. 

I think your solution is already perfect, just provide another solution for you to check. 😄 

Furthermore, I find that the total page number on the navigation bar is always 1 less than the actual one, and I fix it by adding 1. 😄

### DIFF
--- a/packages/client/internals/MenuButton.vue
+++ b/packages/client/internals/MenuButton.vue
@@ -22,7 +22,7 @@ onClickOutside(el, () => {
 
 <template>
   <div ref="el" class="flex relative">
-    <div :class="{ disabled }" @click="value = !value" @keyup.space.prevent>
+    <div :class="{ disabled }" @click="value = !value">
       <slot name="button" :class="{ disabled }" />
     </div>
     <div
--- a/packages/client/internals/MenuButton.vue
+++ b/packages/client/internals/MenuButton.vue
@@ -22,7 +22,7 @@ onClickOutside(el, () => {
 
 <template>
   <div ref="el" class="flex relative">
-    <div :class="{ disabled }" @click="value = !value">
+    <div :class="{ disabled }" @click="value = !value" @keyup.space.prevent>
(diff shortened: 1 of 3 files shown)


antfu:
Thanks for looking into, but actually we could have a better solution for that. 

Just adding 'BUTTON' and 'A' to the list should work
https://github.com/slidevjs/slidev/blob/d9314e72aa6b01be7ea2a4387b396c7ba9e94222/packages/client/state/index.ts#L18-L18

FyisFe:
Have done that. Thanks so much for your reply.

packages/client/internals/MenuButton.vue
```diff
@@ -22,7 +22,7 @@ onClickOutside(el, () => {
 
 <template>
   <div ref="el" class="flex relative">
-    <div :class="{ disabled }" @click="value = !value">
+    <div :class="{ disabled }" @click="value = !value" @keyup.space.prevent>
```
antfu, line 25: Could convert this to `<button/>` I think

packages/client/internals/NavControls.vue
```diff
@@ -7,19 +7,15 @@ import { configs } from '../env'
...
     default: 'fixed',
   },
 })
-
 const md = breakpoints.smaller('md')
 const { isFullscreen, toggle: toggleFullscreen } = fullscreen
-
 const presenterLink = computed(() => `${location.origin}/presenter/${currentPage.value}`)
 const nonPresenterLink = computed(() => `${location.origin}/${currentPage.value}`)
-
```
antfu, line 22: Can you bring them back? Thanks

FyisFe: Sure. Sorry for not noticing them 😅

PR #55, opened 2021-05-10
docs: add guide to use with Docker

Stig124:
This PR adds a Dockerfile and docs update to use docker image, which is up for public grab in docker hub ([url](https://hub.docker.com/r/stig124/slidev)) *(didn't have time to make a README over there)*

*Tell me if you want write permissions on the Docker repo, or even moving it to GHCR in this repo*


entrypoint.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/sh
+
+if [ -z "${KEEP}" ]; then
```
lodenrogue, line 3: What is KEEP variable?

lodenrogue (approved):

Stig124:
> As a non Docker user, this looks quite complex to me 😅 Can you maybe explain a bit why we need a docker for slidev? To me, `npm init slidev` is much simpler and also faster (i suppose)

Mainly dependencies mess : many people don't have the time or can't install dependencies for a quick run, they can pull the image and run the container without any leftover on the host in a very short amount of time

It can also be used to be deployed many presentations at a time, as it can be easily ran in the background by adding a `-d` command switch to the command

And also it can resolve the port issues, some people cannot run (mainly because port is already in use) the app at default port, this will route out the app at any port that the person has permission to bind

SaekiRaku:
Supplement: This looks like another way to solve https://github.com/slidevjs/slidev/issues/63 (For peoples who already have the docker on theirs machine), and have the least code influence. By putting all dependencies in to the docker image, It seems that I can quickly start to write slides with only a single Markdown file by this command (But I still think install Slidev globally is better :D):

```
docker run -d -v /path/to/slides.md:/WORKING_DIR/slides.md -p  slidev@latest
```

antfu:
Adding the docs / guides for using Docker would be good to have, thank you. However, I don't want to turn into this repo into a docker image source at this moment as I will not be able to maintain them either way. So maybe let's leave `Dockerfile` and `entrypoint.sh` for now and do the documentation only. Also kinda feel it's better to use npm to create the docker image as the main branch are not always stable. I wouldn't mind to point your repo and images in the docs atm. WDYT? Thanks.

Stig124:
> Also kinda feel it's better to use npm to create the docker image as the main branch are not always stable

It already uses only npm to build, I even have a [warning](https://github.com/slidevjs/slidev/pull/55/files#diff-c749a8248e06dd9d3f555251b87a61d355c79c44f4df0123837be8da9838b7a4R45) in the docs about this

This has never been what the Docker image is made for
> I will not be able to maintain them either way

I will be able to maintain it without issue, I have a script running which uses [gitpunch](https://github.com/vfeskov/gitpunch) to auto rebuild at each releases on npm and push the docker image to hub

> However, I don't want to turn this repo into a docker image source at this moment

Some people want to build their Docker images from source..., like i said I will be able to maintain it build-side.
 And the files won't need any changes in the future (only maybe a version bump) it can stay "frozen" on this repo, everything happens build-side

docs/guide/install.md
```diff
@@ -37,3 +37,61 @@ $ npx slidev
 > ```bash
 > echo 'shamefully-flatten=true' >> .npmrc
 > ```
+
+## Install on Docker
+
+If you need a rapid way to run a presentation, you can use the prebuilt (or even build your own) [docker](https://hub.docker.com/r/stig124/slidev) image (maintained by [stig124](https://github.com/Stig124))
+
+##### *(this is not meant as a developper environnement, as it uses only npm-fetched binaries)*
```
antfu, line 45:
```suggestion
##### *this is not meant to be a developing environment, it uses only npm-fetched distributions*
```

docs/guide/install.md
```diff
@@ -37,3 +37,61 @@ $ npx slidev
...
+
+```bash
+$ docker pull stig124/slidev:latest
+```
+
+> [podman](https://podman.io/) users can pull the image using the complete URL
+>
+> ```bash
+> $ podman pull docker.io/stig124/slidev:latest
+> ```
```
antfu, line 57: It looks like a more opinionated tools, which users could choose whatever they prefer, I don't think we need to document this here.

docs/guide/install.md
```diff
@@ -37,3 +37,61 @@ $ npx slidev
...
+
+```bash
+$ docker run --rm -v .:/root/slides -p 3000:3030 stig124/slidev:latest
+```
+
+> You can change the port that will face outside by changing the number left of the port declaration => `3000:3030`
+
+Open a browser on `localhost:3000` and you're ready to go, you can leave at any time using `Ctrl` + `C`
+
+> Please note that if you have a `node_modules` folder within the folder with the `slides.md` file, It will be **deleted** *(You can add `-e KEEP=1` next to `--rm` in the command to revert this behaviour)* as it may blocks the starting of the container
```
antfu, line 70: Why do we need to delete it? Is that possible to not mount it?

docs/guide/install.md
```diff
@@ -37,3 +37,61 @@ $ npx slidev
...
+
+### Build the docker image
+
+If you want to build the image by yourself, you can :
```
antfu, line 74:
```suggestion
If you want to build the image by yourself from source, you can:
```

docs/guide/install.md
```diff
@@ -37,3 +37,61 @@ $ npx slidev
...
+
+Clone the repo
+
+```bash
+$ git clone https://github.com/slidevjs/slidev.git
+```
+
+> As the image downloads `slidev` from npm, you can also just download the `Dockerfile` and `entrypoint.sh` from the `Github` website
+
+Go to the folder where you have the `Dockerfile` and `entrypoint.sh` file
```
antfu, line 84: Feels a bit unnecessary as we just have them in the root. could just have this tree commands

```bash
$ git clone https://github.com/slidevjs/slidev.git
$ cd slidev
$ docker build -t <tag> .
```

docs/guide/install.md
Stig124: podman is the drop-in replacement to Docker, deprecating Docker on some Enterprise Linux

docs/guide/install.md
Stig124: This is not a build from source, just the Dockerfile is

docs/guide/install.md
antfu: like `yarn` to `npm`?

docs/guide/install.md
antfu: Ah, then it's might not need to clone the main repo, otherwise this could introduce some confusing, as the npm installs might behavior differently than local source. Or we could move the script to another repo in the org that just contains those files.

Stig124: We can

docs/guide/install.md
Stig124: I don't know the JS world, but pretty much yeah

antfu: Then I'd prefer just simply:

~~~md
```bash
$ docker pull stig124/slidev:latest
```

Or

```bash
$ podman pull docker.io/stig124/slidev:latest
```
~~~

docs/guide/install.md
antfu: Alright, what the common name of it? `slidevjs/dockfile`? I can create the repo and invite you

entrypoint.sh
antfu: Maybe `KEEP_NODE_MODULES`?

docs/guide/install.md
Stig124: I would prefer `container`, but `docker` or `image` would be a good name

antfu: https://github.com/antfu/container Here you go

antfu: Here you go https://github.com/slidevjs/container

docs/guide/install.md
```diff
@@ -37,3 +37,57 @@ $ npx slidev
 > ```bash
 > echo 'shamefully-flatten=true' >> .npmrc
 > ```
+
+## Install on Docker
+
+If you need a rapid way to run a presentation, you can use the prebuilt (or even build your own) [docker](https://hub.docker.com/r/stig124/slidev) image (maintained by [stig124](https://github.com/Stig124))
```
antfu, line 43: Thanks for managing the migrations. Since we already have the repo with the readme, how about let's keep this section with a link to the repo so we don't need to maintain duplicated docs?

Stig124: And there is done

antfu:
Did a bit rephrasing. Thanks a lot on this!

tangramor:
I created a new docker image for dev & deploy slidev pages

# Slidev Docker image

Work with [Slidev](https://sli.dev/). Just run following command in your work folder:

```bash
docker run --name slidev --rm -it \
    --user node \
    -v ${PWD}:/slidev \
    -p 3030:3030 \
    tangramor/slidev:latest
```

If your work folder is empty, it will generate an template `slides.md` and other related files under your work folder, and launch the server on port `3030`. 

You can access your slides from http://localhost:3030/


## Build deployable image

Or you can create your own slidev project to a docker image with Dockerfile:

```Dockerfile
FROM tangramor/slidev:latest

ADD . /slidev

```

Create the docker image: `docker build -t myppt .`

And run the container: `docker run --name myslides --rm --user node -p 3030:3030 myppt`

You can visit your slids from http://localhost:3030/


## Build hostable SPA (Single Page Application)

Run command `docker exec -i slidev npx slidev build` on the runing container `slidev`. It will generate static HTML files under `dist` folder.

You can host `dist` in a static web site such as Github pages or Gitlab pages. You can also host it by your self:

```bash
docker run --name myslides --rm -p 80:80 -v ${PWD}/dist:/usr/share/nginx/html nginx:alpine
```

Or create a static image with following Dockerfile:

```Dockerfile
FROM nginx:alpine

COPY dist /usr/share/nginx/html

```

Create the docker image: `docker build -t mystaticppt .`

And run the container: `docker run --name myslides --rm -p 80:80 mystaticppt`

You can visit your slids from http://localhost/

PR #90, opened 2021-05-12
docs(themes): add Vercel theme

nico-bachner:
Since the theme is still in development, this PR can also be merged at a later point in time.


nico-bachner:
I see a test didn't pass. Should I restructure the sentence so that there is no apostrophe? That would allow me to use single quotes without breaking the build.

PR #101, opened 2021-05-13
don't alter in code-block style tags

tskvrnda:
I've ran in a problem where I have <style> tag in a code block (e.g. vue), which are not supposed to be altered as they are not part of the slide styles.


antfu:
Parser should not transform the code as the content will also be used to write back. See d2a5b96be5a454a791052815ec5c2369673f1e68

Thanks for the contribution!

PR #112, opened 2021-05-14
docs(syntax): fix the tweet component usage

shoichiaizawa:
With the v-bind `:id` prop, the Tweet component doesn't always work.

e.g.
`@antfu7`'s pinned tweet doesn't render with the prepending colon (:).
`<Tweet :id="1389800180002672641" />` doesn't work
but `<Tweet id="1389800180002672641" />` works.

Also, the closing `div` doesn't have slash, causing a server error.


antfu:
Love how JavaScript work :P 
image

PR #114, opened 2021-05-16
feat: shortcuts auto repeating

Marr11317:
Currently, keeping right pressed does nothing more than just pressing it once. The event should auto repeat if the key combination is still pressed. 
This PR is an untested work. I have not built slidev, I just wanted to try it out and ask if you thought it was a good idea. I actually modified the code in github's default, so it might not even build. If you think that the idea is worth a bit of work I will put it.


Marr11317:
Great to see there was some interest;)

PR #140, opened 2021-05-20
fix: cli prompts

HomyeeKing:
for me, at the first time I create the **slidev** app

at the step of giving `project name`, the `slidev` is default (actually I think  that should be put  at the question) 

 if you want to rename it , just type your name directly, but I didn't know it firstly and kept typing `backspace`  😅

in this operation, I typed two `space` unexpectedly and I didn't notice it, so the generated directory is with the `space` prefix

that's why I can't  `cd` into it 😅

### DIFF
--- a/packages/create-app/index.js
+++ b/packages/create-app/index.js
@@ -32,10 +32,9 @@ async function init() {
     const { projectName } = await prompts({
       type: 'text',
       name: 'projectName',
-      message: 'Project name:',
-      initial: 'slidev',
+      message: 'Project name: (slidev)',
     })
-    targetDir = projectName.trim()
+    targetDir = projectName.trim() || 'slidev'
   }
   const packageName = await getValidPackageName(targetDir)
   const root = path.join(cwd, targetDir)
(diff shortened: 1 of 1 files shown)


antfu:
I don't think you need to `backspace` the placeholders, you can directly type the name you want. I'd prefer to keep the original usage of `initial` options. But I do agree to trim the result.

HomyeeKing:
yes, but it's not an intuitive operation I think
only if you have tried 1、2 times then you know, oh I just need to type name directly 
people like me still want to delete it first subconsciously even have known about that


most of users may need to  replace by their own project name, so why not just leave a blank to let them fill in
and the default name is in the question part, just skip by `Enter`  if you  don't

antfu:
Probably better to raise this discussion in `prompts`? The placeholder has grayed out, which is intuitive enough for me. This could be a subjective preference, and I don't see it's worth it to implement our own since prompts already provided a solution for that.

HomyeeKing:
it depends on the theme of your terminal, for me like this
image


it does a subjective preference, so I'll revert the last commit

PR #157, opened 2021-05-22
fix: windicss extraction, vue import, demo text color

hannoeru:


antfu:
Looks good, what's the obstacle?

PR #161, opened 2021-05-22
fix: windicss extrant target in monorepo

hannoeru:


packages/client/package.json
```diff
@@ -31,8 +31,12 @@
     "nanoid": "^3.1.23",
     "prettier": "^2.3.0",
     "recordrtc": "^5.6.2",
+    "resolve": "^1.20.0",
     "vue": "^3.0.11",
     "vue-router": "^4.0.8",
     "windicss": "^3.0.9"
+  },
+  "devDependencies": {
+    "@types/resolve": "^1.20.0"
```
antfu, line 40: Let's move this to top level devDeps

hannoeru:
Looks like it braked something...

antfu:
Should be fixed now

PR #162, opened 2021-05-22
chore: use resolve pkg to resolve import path

hannoeru:


antfu:
Confused, how about do it with #161 in one-go? Otherwise they will be conflicted anyway

hannoeru:
ok, I will move it to #161

PR #164, opened 2021-05-22
fix: exclude node_modules but keep @slidev/client in windicss scan

hannoeru:


antfu:
Have you tested it? I am not sure if we support `!` for globs 🤔

hannoeru:
I tested it, the package used in vite-plugin-windicss support that syntax.

antfu:
Alright, I trust you then

PR #172, opened 2021-05-24
fix: filter head title when it is styled using html tags

dizys:
This PR filters unescaped HTML tags in the head title.

## Background

Currently, head title can be styled using HTML tags:

For example,

```markdown
# <span style="color:red">Composable</span> Vue
```

However, because it is also directly fed to the browser tab title like this:
https://github.com/slidevjs/slidev/blob/3afc764ffa628f06d7bd11fbd4f5a751e3777e9a/packages/client/setup/root.ts#L18-L20

So the literals would straight show up in the browser tab, being something like `<span style="color:red">Composable...`, which is obviously not very ideal.

## Solution

This PR utilizes the regular expression to filter unescaped HTML tags in the head title when calling `useHead`.

The regular expression being:
```ts
const REGEX_HTML_TAGS_IN_MARKDOWN = /(?<!\\)<(.*?)[^\\]>/g
```

Note that escaping using slashes would be allowed by this regular expression. For instance, markdown `\<span style="color:red"\>` would mean literal `<span style="color:red">`.

### DIFF
--- a/packages/client/setup/root.ts
+++ b/packages/client/setup/root.ts
@@ -7,6 +7,7 @@ import { router } from '../routes'
 import configs from '/@slidev/configs'
 // @ts-expect-error
 import serverState from '/@server-ref/state'
+import { stripMarkdownHTMLTags } from '../utils'
 
 export default function setupRoot() {
   // @ts-expect-error
@@ -16,7 +17,7 @@ export default function setupRoot() {
   /* __injections__ */
 
   useHead({
-    title: configs.title ? `${configs.title} - Slidev` : 'Slidev',
+    title: configs.title ? `${stripMarkdownHTMLTags(configs.title)} - Slidev` : 'Slidev',
   })
 
   // sync with server state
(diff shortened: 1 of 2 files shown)


dizys:
Having second thought about using regex lookbehind. This can cause serious incompatibility issues for Safari users. 🤔

antfu:
Not sure, probably better to specify the title yourself explicitly:

```md
---
title: Whatever you want
---

# <b>Whatever</b>
```

dizys:
> Not sure, probably better to specify the title yourself explicitly:
> 
> ```md
> ---
> title: Whatever you want
> ---
> 
> # <b>Whatever</b>
> ```

Yeah, that seem to make the most sense for now. Tho, having it filtered automatically would be a smoother experience out of the box, imo.

I'll leave it as it is now, may come back to this pr later. If you think this is unnecessary, please go ahead and close this PR, it would be no problem at all.

👏This project is a brilliant idea by the way.

antfu:
Yeah, this could be overkill when time I might just want to have html tags or type genetics in the title. Being able to specify the title could be better to me.

Meaning:

```
# Talk about `<details/>`
```

```
# What is Readonly<T> in TypeScript
```

Thanks for the contributions anyway!

dizys:
> ```markdown
> # What is Readonly<T> in TypeScript
> ```

Just to be clear, in your second example, I think you may have to do escaping like `# What is Readonly\<T\> in TypeScript` to be able to properly render it.

PR #174, opened 2021-05-25
fix: Non-number slideScale causing mispositioning

### LINKED ISSUE #173: Slide appears in bottom right corner
**Describe the bug**
Slides appear in the bottom right corner after a refresh. If I click either of the `scale` options, they are back where they belong. I can also fix by clearing localstorage. With either fix, the issue re-appears after a couple of refreshes.

**To Reproduce**
1. `npm init slidev@latest`, install and start now, npm. Browser will open to `localhost:3030` automatically.
2. Refresh the page. Observe that the slides are now in the bottom right corner.

video demo: https://user-images.githubusercontent.com/2491257/119438940-241f8200-bcd6-11eb-8e4b-49fa599a6e31.mp4

 - OS: MacOS Mojave 10.14.6
 - Browser: Chrome Version 90.0.4430.93 (Official Build) (x86_64)
 - Slidev version: v0.13.10

dizys:
Fix #173 / #177.

Using `null` in number typed `slideScale` would also make Vue runtime type checking fail.

But notice this fix gives special meaning to value `0`, replacing the original `null`.

### DIFF
--- a/packages/client/internals/Settings.vue
+++ b/packages/client/internals/Settings.vue
@@ -3,10 +3,10 @@ import { slideScale } from '../state'
 import SelectList from './SelectList.vue'
 import type { SelectionItem } from './types'
 
-const items: SelectionItem<number | null>[] = [
+const items: SelectionItem<number>[] = [
   {
     display: 'Fit',
-    value: null,
+    value: 0,
   },
   {
     display: '1:1',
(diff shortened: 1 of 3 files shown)


dizys:
I've tried replacing `null` with `NaN` instead, but `NaN` would require special comparison `isNaN` in order for `SelectList` in `Settings.vue` to work propertly with the checkmark class. So `0` is my best attempt so far.

PR #199, opened 2021-05-30
fix: info modal is not shown in presenter mode

tmkx:


antfu:
Maybe it's better to hide the info button on Presenter mode?

tmkx:
> Maybe it's better to hide the info button on Presenter mode?

Done ; )

PR #203, opened 2021-05-31
fix: remove html tags in title

tmkx:
Fix bugs in the following screenshot:

image


tmkx:
https://github.com/slidevjs/slidev/pull/172

tmkx:
Maybe we can reconsider this case. If we want to use html tags in title, it's better to escape it.

image

antfu:
Alright, I like the approach, let's have it, thanks!

PR #210, opened 2021-06-01
Plotly asset

jorekai:
Hey thought I'd publish a small component I used for my thesis' presentation. It supports a fullscreen Plotly embedded widget by wrapping the iframe like the youtube component. Proceed how you want to :), maybe others will need it too.

Usage:

1.  Create chart [here](https://chart-studio.plotly.com)
2.  Make use of it! (edit: set slide _layout: none_)
    ```vue
    <Plotly link="//plotly.com/~jorekai/3.embed"/>
    // setting width and height is not recommended
    <Plotly link="//plotly.com/~jorekai/3.embed"/ width="100px" height="100px">
    ```

### DIFF
--- a/packages/client/builtin/Plotly.vue
+++ b/packages/client/builtin/Plotly.vue
@@ -8,12 +8,12 @@ Usage:
 -->
 
 <script setup lang="ts">
-import { defineProps } from "vue";
+import { defineProps } from 'vue'
 defineProps<{
-  link: string; // plotly link for chart
-  width?: number;
-  height?: number;
-}>();
+  link: string // plotly link for chart
+  width?: number
+  height?: number
+}>()
 </script>
 
 <template>
--- a/packages/client/builtin/Plotly.vue
+++ b/packages/client/builtin/Plotly.vue
@@ -1,0 +1,30 @@
+<!--
+Wrapper for embedded plotly charts - export your chart from chart-studio
+
+WARNING: Layout of page must be set to none, charts only work in fullscreen mode reliably
+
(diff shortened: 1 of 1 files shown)


dizys:
Great seeing someone integrating Plotly here.

Somehow I think a more universal iframe component would be more suitable to be built-in? Call it a `<Browser>` or something, cause this is basically what it does.

I'm also figuring out a more convenient way to use a Vue component installed directly from `yarn add ...`, maybe you can consider publishing the vue component as an npm package? Just a thought.

I really don't know how much and to what extent should Slidev include 3rd party components as built-ins. It would be fantastic if we have a curated Slidev components registry where we can just find the component, `yarn add` and integrate without much configuration.

antfu:
Thanks for the contributions. But as this is basically a thin wrapper for iframe, I don't really see much value to introduce a built-in component for it (you can either use iframe directly in markdown, or create components under your local `./components`).

For now, I'd like to keep Slidev's built-in components simple and for more general and common usages. Maybe later we could consider having a shared components lib as @dizys proposed.

jorekai:
Hey completely understandable decision. I like the idea of slidev providing some kind of additional, loosely coupled library. This would improve the slide creation speed in general. So @dizys approach also makes much sense to me.

Especially with clear contribution guidelines and snapshot testing this would invite maintainable contributions. If you are thinking about initializing such a library I'd definitely support it myself. I think you will get many more contributors on your project as time goes by.

I personally enjoyed using this over power point so much, keep up the good work :)

PR #218, opened 2021-06-04
Add option "align" for Mermaid diagrams

christian-nils:
Hello,

I have created this pull request because I needed a way to center my diagrams when a scaling was applied. This creates the parameter "align" in the mermaid MD markup. The options are: 'left', 'center', and 'right'.

So instead of:

```markdown
    ```mermaid {theme: 'neutral', scale: 0.8}
    graph TD
    B[Text] --> C{Decision}
    C -->|One| D[Result 1]
    C -->|Two| E[Result 2]
    ```
```

You can write:

```markdown
    ```mermaid {theme: 'neutral', scale: 0.8, align: 'center'}
    graph TD
    B[Text] --> C{Decision}
    C -->|One| D[Result 1]
    C -->|Two| E[Result 2]
    ```
```

This is quite useful when you want to display a single centered (scaled) diagram on a slide.

Cheers,

Christian-Nils

### DIFF
--- a/packages/client/builtin/Mermaid.vue
+++ b/packages/client/builtin/Mermaid.vue
@@ -25,7 +25,6 @@ const props = defineProps<{
 
 const vm = getCurrentInstance()
 const html = computed(() => renderMermaid(props.code || '', Object.assign({ theme: props.theme }, vm!.attrs)))
-
 </script>
 
 <template>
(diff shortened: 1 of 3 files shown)


christian-nils:
Note that the choice of the parameter was poor since it conflicts with the align attribute. We might want to consider renaming the parameter.

antfu:
Should be covered by #225 right? Thanks!

Closing

PR #225, opened 2021-06-06
feat: apply real resize to mermaid graphs

christian-nils:
Hi,

Here comes another way to scale the SVG generated by MermaidJS. As it is done now, the Transform component wrapping the Mermaid component makes it difficult to position the generated SVG when scaled.

I proposed a simple addition (https://github.com/slidevjs/slidev/pull/218) to expose the alignment option to the user. But it may still be unsufficient in some cases (vertical adjustments for instance)

The present PR proposes a way to actually scale directly the SVG and remove the Transform component. The generated SVG become then much easier to position.

Feel free to reject both PRs if this does not make any sense. (I personally prefer this second PR)

Have a good day!

/CN


antfu:
Typo fixed, thanks

PR #228, opened 2021-06-07
feat: shortcuts configuration

### LINKED ISSUE #209: Configure navigation shortcuts
When I used slidev click animation, I feel confused with shortcuts between shortcuts left/right and up/down, which often leads me to mis-touch. And it is inconvenient for me that there is no shortcut bindings on enter and backspace.

I think it is necessary to add some support for the configuration of shortcuts for some users in need.

1. export some built in navigation functions
2. add a new type of setup script, like `defineShortcutsSetup`

yjl9903:
Fixes #209 

Add shortcuts configuration.

Create `./setup/shortcuts.ts` with the following content:

```ts
import { defineShortcutsSetup, NavOperations } from '@slidev/types'

export default defineShortcutsSetup((nav: NavOperations) => {
  return [
    {
      key: 'enter',
      fn: () => nav.next(),
      autoRepeat: true,
    },
    {
      key: 'backspace',
      fn: () => nav.prev(),
      autoRepeat: true,
    },
  ]
})
```

Related type declaration:

```ts
export interface NavOperations {
  next: () => void
  prev: () => Promise<void>
  nextSlide: () => void
  prevSlide: () => Promise<void>
  downloadPDF: () => Promise<void>
  toggleDark: () => void
  toggleOverview: () => void
  escapeOverview: () => void
  showGotoDialog: () => void
}

export interface ShortcutOptions {
  key: string
  fn: () => void
  autoRepeat?: boolean
}

export type ShortcutsSetup = (nav: NavOperations) => Array<ShortcutOptions>

export function defineShortcutsSetup(fn: ShortcutsSetup) {
  return fn
}
```

### DIFF
--- a/packages/client/logic/shortcuts.ts
+++ b/packages/client/logic/shortcuts.ts
@@ -36,44 +36,32 @@ export function strokeShortcut(key: KeyFilter, fn: Fn) {
 }
 
 export function registerShortcuts() {
-  // user custuom shortcuts
-  const custuomShortcuts = setupShortcuts()
+  const customShortcuts = setupShortcuts()
 
-  // default global shortcuts
-  new Map<string, ShortcutOptions>([
-    { key: 'space', fn: next, autoRepeat: true },
-    { key: 'right', fn: next, autoRepeat: true },
-    { key: 'left', fn: prev, autoRepeat: true },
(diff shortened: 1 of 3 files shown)


antfu:
Could you also open another PR to the docs repo adding these info? Thanks!

PR #237, opened 2021-06-08
fix: use viewBox attribute instead of height when scaling Mermaid diagrams

christian-nils:
Hi,

In the new scaling method introduced earlier, the mermaid SVG's original `height` attribute is used. However, in some cases, the SVG returned my mermaidjs has a `height` attribute equal to "100%" which makes the scaling non consistent.

Using the viewBox attribute of the SVG should solve this issue, since `height/viewBox[3]` can directly be interpreted as a scaling factor for the SVGs.

Thanks!

/CN

### DIFF
--- a/packages/client/builtin/Mermaid.vue
+++ b/packages/client/builtin/Mermaid.vue
@@ -13,7 +13,6 @@ pie
 -->
 
 <script setup lang="ts">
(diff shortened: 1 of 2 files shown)


antfu:
Nice fix! Thanks

PR #269, opened 2021-06-21
feat(cli, client): export pages for every clicks

jk4235:
export add a flag `everyClicks`
use `slidev export --everyClicks` export pages for every clicks


antfu:
Thanks! For the record, I have rename to option to `--with-clicks` instead.

PR #274, opened 2021-06-22
fix: cannot read property 'replace' of undefined

WingLim:
configs.titleTemplate maybe undefined if user didn't set it.


antfu:
It shouldn't be undefined. What's your usage?

WingLim:
Oh, I found why it's undefined for me, I am using `@slidev/cli v20.5` to run `pnpm run demo:dev`.
It's stupid...

Issue #217, opened 2021-06-03
`v-click`, `v-after`, and `v-click-hide`


ksnyde:
## Description

I find the description of `v-link` fine and intuitive but after that it goes down hill pretty quickly and there are no examples to go in the docs or the test skeleton you get when you create. I did try to use `v-after` but must admit it still made no sense to me.

More fundamentally I think there is an attempt to treat click transitions as a 2-state problem when in fact it is at a minimum a 3 state problem. The states are:

> Note: I am ignoring the _initial_ state when a user first enters the slide; this might be considered too but it probably better left as a "non-click block" and therefore not a part of scope of this feature request directly

0. [ unclicked ] The click blocks which are not yet "in scope" (which typically means _not yet visible_)
1.  [ current ] The most recent "click" block
2.  [ clicked ] The click blocks which have been brought into scope but are not the "current" item

Currently state **clicked** and **current** are represented by CSS classes but not the **clicked** state. I believe it is _very_ desirable to distinguish between **current** and **clicked** for two key reasons:

1.  **Focus:** Presenters often want to "dim" prior clicked states so that the focus of the audience can be maintained 
2. **Space Management:** sometimes you just want the prior blocks to "disappear"; the advantage being that you can reclaim the vertical space

### Preamble
As I am currently a bit unclear on the intent of `v-after` and `v-click-hide`, I may be missing features that this solution does not provide but my goal is to reduce the primitives down to just one: `v-click`.

### Proposed changes to `v-click`

1. Allow for indexation with property on `v-click`:

    ```html
    <v-click index="1" />
    ```

    Allows you to state the step count in which the given block will be brought into scope. This is intended to work with the functionality exposed in `<v-clicks>` but for a single block element instead of a list.

2. Provide `hide` boolean flag:

    ```html
    <v-click hide />
    ```

    Inverts the behavior of the block. Instead of starting out hidden, it starts out visible but then becomes hidden when clicked.

3. Provide `after` class prop:

    ```html
    <v-click after="text-gray-300" />
    ```

    Provides an alternate way of styling the different click states that matches up well with WindiCSS's super powers while at the same time providing a shorthand for styling at the _per block_ level (this can often be desirable)

4. Provide `current` and `before` class props:

    ```html
    <v-click before="opacity-10" current="opacity-100" after="opacity-50" />
    ```

    Kind of self-explanatory at this point. I called out `after` first because I think it would end up being the most used but all three could be very useful.

[...]

PR #277, opened 2021-06-22
feat(v-click): use order map to order v-click

WingLim:
As I said in #272 .

I add a order map to record `v-click` order.

`orderMap` struct is:

```ts
orderMap: Map<number, HTMLElement[]>
```

Key is order `number`
Value is `HTMLElement` array

For `v-click` which not set their order, they will set their order as their index in `elements`.

The reason for using array as map value is for `v-after`.

When initialize one `v-after` element, it will push itself to the last of array.

Then, we can set the last item of array to `current` state, the others items to `prior` state.

This implement is better than the #271, it looks more elegant.

For simplfy `v-click-hide`, I add `hide` prop for `<v-click />`.

And I think it will be better the `v-click-hide` just call some thing like `<v-click hide />`, but don't know how to implement it.

Still, the state design comes from #217

### DIFF
--- a/cypress.json
+++ b/cypress.json
@@ -1,3 +1,4 @@
 {
-  "baseUrl": "http://localhost:3030"
+  "baseUrl": "http://localhost:3030",
+  "chromeWebSecurity": false
 }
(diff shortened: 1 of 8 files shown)


antfu:
Awesome feature! Thank you!

Issue #310, opened 2021-07-18
Line number in code block 


mathieutu:
**Is your feature request related to a problem? Please describe.**
Hi, thanks for this cool project!

When explaining pieces of code to an audience it's really useful to be able to tell them that "you can find xxx at line number yyy", and for them it's easier for asking question.

It would also be useful to have the line numbers for click highlighting (I count lines all the day to know which one is what)

**Describe the solution you'd like**
As usual, a simple column at the beginning of the block.

**Describe alternatives you've considered**
😶

Thanks!

I have no time for this right now, but I could try a PR to add it!

antfu:
Happy to have it, PR welcome!

(Prism seem to easier to do, I don't mind to have the features only available on one highlighter first, so you don't need to take the task too big. For Shiki, https://github.com/shikijs/shiki/issues/3

mathieutu:
Hi, 
I'm currently using Shiki, so I've put in #311 something basic I've quickly made on my own project, based on https://github.com/shikijs/shiki/issues/3#issuecomment-830564854.

antfu:
#311

fr0zn:
Would it be possible to provide a way to set the initial line number for each code block and allow line highlighting based on the new numbers?

PR #311, opened 2021-07-18
feat: support line numbers for code blocks

mathieutu:
Quickly and basically done, based on https://github.com/shikijs/shiki/issues/3#issuecomment-830564854

see https://github.com/slidevjs/slidev/issues/310


antfu:
Update the implementation to have general line number support and configurable via an option. Nice solution you have found. Thank you!

PR #342, opened 2021-09-12
fix: catch cli open shortcut errors

mnixry:
Sometimes, the `slidev --open` shortcut command fails (such as browser not exist), and it causes the entire development server to fail.
I'm trying to add a catch statement to suppress this situation

### DIFF
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -158,7 +158,7 @@ cli.command(
             sh.action()
           }
           catch (err) {
-            console.error(`error occurred while executing shortcut ${sh.fullname}: ${err}`)
+            console.error(`Failed to execute shortcut ${sh.fullname}`, err)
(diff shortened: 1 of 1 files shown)


packages/slidev/node/cli.ts
```diff
@@ -149,9 +152,15 @@ cli.command(
...
-        const sh = SHORTCUTS.filter(item => item.name === str)[0]
-        if (sh)
-          sh.action()
+        const [sh] = SHORTCUTS.filter(item => item.name === str)
+        if (sh) {
+          try {
+            sh.action()
+          }
+          catch (err) {
+            console.error(`error occurred while executing shortcut ${sh.fullname}: ${err}`)
```
antfu, line 161:
```suggestion
            console.error(`Failed to execute shortcut ${sh.fullname}`, err)
```

PR #361, opened 2021-09-30
feat: introduce frame layouts

### LINKED ISSUE #360: frame layouts
One of the key benefits of using slidev is being able to directly include web-demo's in my slides. This obviates the need to screenshot everything, which gives a very static impression. 

For instance we can include live documentation in the presentation etc. [Other users](https://quantumblacklabs.github.io/kedro-community/7) have found the same.

It surprised me that there is not yet a default layout for that case.

Layouts analogously to image where we provide a web page out-of-the-box.

It's trivial to extend slidev (hackable)! If we decide it's not general enough, then users could extend it themselves. For the reason give in the first section, I would argue to include it by default.

sbrugman:
Closes #360

Docs PR linked.

### DIFF
--- a/packages/client/layouts/iframe-left.vue
+++ b/packages/client/layouts/iframe-left.vue
@@ -1,0 +1,23 @@
+<script setup lang="ts">
+import { defineProps, ref } from "vue";
+const props = defineProps({
+  class: {
+    type: String,
+  },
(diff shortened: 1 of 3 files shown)


antfu (approved):
Cool, I like it!

PR #382, opened 2021-10-21
Added a center-top, bottom-left-right layout 

Leone25:
Idk really what to name this but I though this would be useful for other people so here we go


antfu:
Thanks for the PR. But since it's a fairly simple layout to do, I suggest keeping it in the user land to provide more flexibility.

Leone25:
alright i guess :L

PR #451, opened 2022-01-12
feat: add Toc component

tonai:
Adding a built-in TOC component.

TOC is really useful when you have long presentations (like for a 3 days training for example). It helps people not to be to much lost.
Lots of presentation library miss this feature, that is mandatory in the usage I have.
I found this library awesome, that's why I am proposing this feature, because It would even more incredible with this feature :smiley: 

This feature automatically gather informations about the title and the title level of each slides to build a tree.
You can also set both information using front matter syntax (if your slide does not contain a title, or if you want to override it).
For example:
```md
---
title: Amazing slide title
title level: 2
---
```

Available props for the Toc component:
* columns (default = 1): TOC in multiple columns
* maxDepth (default = Infinity): Maximum depth of the TOC
* minDepth (default = 1): Minimum depth of the TOC
* mode (default = 'all'):
  * 'all': Displays all items
  * 'onlyCurrentTree': Displays only items that belong to the current active slide (active slide, parents of active slides and children of active slide)
  * 'onlySiblings': Displays only items that belong to the current active slide and their siblings (but it will not show the children of a slide that does not belong to the current tree)

You can also mark a slide to not be displayed in the TOC with the front matter syntax:
```md
---
skipInToc: true
---
```

Examples (Table of content slide is skipped):

=> `columns="3"`
screencapture-localhost-3031-2-2022-01-12-16_59_50

=> `maxDepth="2"`
screencapture-localhost-3030-12-2022-01-12-17_03_38

=> `minDepth="2"`
screencapture-localhost-3030-21-2022-01-12-17_04_04

=> `mode="onlyCurrentTree"`
screencapture-localhost-3031-17-2022-01-12-17_00_36

=> `mode="onlySiblings"`
screencapture-localhost-3031-31-2022-01-12-17_01_15

Possible enhancements:
* add some styles to see active item in the list (classes are already added but I am not sure where I should put the styles or if we should let the user add his own styles)
* the tree for the TOC is a computed value inside the Toc component, but if you put multiple Toc component in your presentation, the tree get calculated for each components. Maybe we can calculate the tree globally and get it inside the component (that will filter it accordingly to the props), but I am not sure how we should do that ().

### DIFF
--- a/packages/client/builtin/Toc.vue
+++ b/packages/client/builtin/Toc.vue
@@ -1,102 +1,117 @@
 <!--
 Table Of content
 
+`mode` can be either 'all', 'onlyCurrentTree' or 'onlySiblings'
+
 Usage:
 
-<Toc columns="2" maxDepth="3" mode="onlySiblings"/>
+<Toc columns='2' maxDepth='3' mode='onlySiblings'/>
 -->
-<script lang="ts">
+<script setup lang='ts'>
+import { computed } from 'vue'
 import type { RouteRecordRaw } from 'vue-router'
-import { defineComponent } from 'vue'
+import { currentRoute, rawRoutes } from '../logic/nav'
 import type { TocItem } from './TocList.vue'
 
-export default defineComponent({
-  props: {
-    columns: {
-      default: 1,
-    },
-    maxDepth: {
-      default: Infinity,
-    },
-    minDepth: {
-      default: 1,
-    },
-    mode: { // 'all' | 'onlyCurrentTree' | 'onlySiblings'
-      type: String,
-      default: 'all',
-    },
-  },
-  computed: {
-    toc() {
-      // console.log(JSON.parse(JSON.stringify(this.$slidev.nav)))
-      let tree = this.makeTree()
-      this.addActiveStatuses(tree)
-      tree = this.filterTree(tree)
-      if (this.mode === 'onlyCurrentTree')
-        tree = this.filterOnlyCurrentTree(tree)
-      else if (this.mode === 'onlySiblings')
-        tree = this.filterOnlySiblings(tree)
-      return tree
-    },
-  },
-  methods: {
-    makeTree(): TocItem[] {
-      return this.$slidev.nav.rawRoutes
-        .filter((route: RouteRecordRaw) => route.meta?.slide?.title)
-        .reduce((acc: TocItem[], route: RouteRecordRaw) => {
-          this.addToTree(acc, route)
-          return acc
-        }, [])
-    },
-    addToTree(tree: TocItem[], route: RouteRecordRaw, level = 1) {
-      const titleLevel = route.meta?.slide?.titleLevel
-      if (titleLevel && titleLevel > level && tree.length > 0) {
-        this.addToTree(tree[tree.length - 1].children, route, level + 1)
-      }
-      else {
-        tree.push({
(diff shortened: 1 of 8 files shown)


packages/client/builtin/Toc.vue
```diff
@@ -0,0 +1,107 @@
+<!--
+Table Of content
+
+Usage:
+
+<Toc columns="2" maxDepth="3" mode="onlySiblings"/>
+-->
+<script lang="ts">
```
antfu, line 8: Could we use the `<script setup>` syntax for overall consistency?

tonai: Of course no problem, but I'm not completely familiar with this new syntax and I don't know how I can access the $slidev inside of it. Can you give me an hint ?

antfu: `getCurrentInstance().proxy.$slidev` I think. But actually you can directly import them from the internal modules

tonai: Thank you for your answer.
I have pushed the update.
Also what is your thought about the last 2 points ?
* add some styles to see active item in the list (classes are already added but I am not sure where I should put the styles or if we should let the user add his own styles)
* the tree for the TOC is a computed value inside the Toc component, but if you put multiple Toc component in your presentation, the tree get calculated for each components. Maybe we can calculate the tree globally and get it inside the component (that will filter it accordingly to the props), but I am not sure how we should do that.

For the last point can I add a computed value for the tree inside the `@slidev/client/logic/nav` ? Will it solves the problem ?

packages/client/builtin/Toc.vue
```diff
@@ -0,0 +1,122 @@
...
+  const titleLevel = route.meta?.slide?.titleLevel
+  if (titleLevel && titleLevel > level && tree.length > 0) {
+    addToTree(tree[tree.length - 1].children, route, level + 1)
+  }
+  else {
+    tree.push({
+      children: [],
+      level,
+      path: route.path,
+      skipInToc: Boolean(route.meta?.skipInToc),
```
antfu, line 45: How about `hideInToc`?

packages/client/routes.ts
```diff
@@ -51,6 +51,8 @@ declare module 'vue-router' {
       id: number
       no: number
       filepath: string
+      title?: string
+      titleLevel?: number
```
antfu, line 55: How about just `level`?

antfu (commented):
Looks great, thanks!

packages/client/builtin/Toc.vue
antfu: 1. I prefer to not have styles, but we could still add classes prefixed by `.slidev` just like other classes for users to override for certain state
2. I don't mind that as I don't think the computed will be very expensive. We could leave it as an future optimization if it really leads to performance issues

tonai:
Thanks for your comments, I fixed them both.
I also moved the computed tree value inside the nav logic file, not really for performance issue, but I think the tree can be reused in ome situation (like for creating a sidebar reported in this issue: https://github.com/slidevjs/slidev/discussions/439)

antfu:
@tonai would be great if you can also help adding this to the docs, thanks!

antfu:
You can add a sub section of something like `### TOC`, and I will fulfill the rest comoponents :)

PR #462, opened 2022-01-18
fix: global layer position

tonai:
According to doc https://sli.dev/custom/global-layers.html

### DIFF
--- a/packages/client/internals/SlidesShow.vue
+++ b/packages/client/internals/SlidesShow.vue
@@ -23,8 +23,8 @@ if (__SLIDEV_FEATURE_DRAWINGS__ || __SLIDEV_FEATURE_DRAWINGS_PERSIST__)
 </script>
 
 <template>
-  <!-- Global Bottom -->
-  <GlobalBottom />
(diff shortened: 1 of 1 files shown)


antfu:
I think the previous behavior is expected as in HTML the later dom have higher z-index

tonai:
Ok then you mean top and bottom relative to the Z axis I suppose.
It was not not really intuitive for me, because top and bottom are most of the time relative to the Y axis.
At first glance I though that global-top would be displayed at the top and global-bottom at the bottom of the slide.
An in fact that's what happen if you create simple components that are not absolutely positionned :
Capture d’écran de 2022-01-18 14-28-26
```vue
<template>
  <div>Global bottom</div>
</template>
```
That's why I find it a little confusiong.

antfu:
I think "Layers" are supposed to be Z-indexed if you ever used Photoshop or similar tools. Feel free to adjust the docs to the statements that more clear to you. Thanks

tonai:
Yes maybe the docs need some clarifications

PR #470, opened 2022-01-31
Allow HTML in TOC

marob:
Typical use-case: having an image in the title.


tonai:
In fact, this is not enough, we should also parse the markdown in the title.
Currently with your PR this will work:
```md
# My title <img alt="image" src="/some_image" />
```
But this will not:
```md
# My title ![image](/some_image)
```

tonai:
I added the markdown to HTML transformation for the slide title.

PR #482, opened 2022-02-15
fix: markdown parsing in toc title

tonai:
I didn't create an issue about that but there is still some issues with the title content.
For exemple if you use an image in the title and then build with the `--base` option, the base won't be injected in the title displayed inside the toc.

To fix this I used the same mechanism you use for building the slides.
I created custom imports:
- `@/slidev/titles`: returns an object matching the slide number and the Vue component to use as title
- `@slidev/title/*.md`: returns the Vue component which will be targeted by the same vite plugins as the slides `@slidev/slides/*.md`

Then the additionnal `div` and `p` get stripped by a new `slidev:title-transform:pre` transformer.

This way we ensure the title used in the TOC will be the same as the title in the slides.

### DIFF
--- a/packages/slidev/node/plugins/loaders.ts
+++ b/packages/slidev/node/plugins/loaders.ts
@@ -249,9 +249,12 @@ export function createSlidesLoader(
 
         // title
         if (id === '/@slidev/titles.md') {
-          return data.slides.map(({ title }, i) => {
-            return `<template ${i === 0 ? 'v-if' : 'v-else-if'}="+no === ${i + 1}">${title}</template>`
-          }).join('')
+          return {
+            code: data.slides.map(({ title }, i) => {
+              return `<template ${i === 0 ? 'v-if' : 'v-else-if'}="+no === ${i + 1}">${title}</template>`
+            }).join(''),
+            map: {},
+          }
         }
 
         // pages


antfu:
I am a bit worried about the extra complexity here. I think this essentially makes the title a full-powered vue component, with the extra cost (payload, perf) of having two components for each slide. To me, it would be a relatively niche case to have such capability in TOC. Can you share more usage of having such a complex TOC?

tonai:
It is a little complex yes, but I don't know an other way to get the same result for the title in the TOC, as in the slides.
If your have any suggestion I am happy to hear them.
In fact I discovered this issue, because I have an use-case were I have an image in the title, and I need to build the presentation using a custom `base`.
But as I said, not having the same workflow for the title as for the slides may results in other bugs in the future that may be hidden for now..

tonai:
Fixing merge conflicts

antfu:
One other idea, instead of serving each title as a virtual component, how about just making `ToC` as a virtual component so you can conact the title's HTML while keeping the component tree simple?

tonai:
> One other idea, instead of serving each title as a virtual component, how about just making `ToC` as a virtual component so you can conact the title's HTML while keeping the component tree simple?

Ok I'll try pushing in that direction

tonai:
In fact I have a little problem with that solution.
Because I also developed a little component for slidev that adds a progress bar here: https://github.com/Smile-SA/slidev-component-progress
And in this component I also want to display the title.
With the solution of the virtual `<toc>` component I won't be able to only use the well parsed title inside this component.
I'll try to search for something that fits both sides.

tonai:
Ok I simplified by having only one virtual component that is `/@slidev/titles.md`.
It's a component that takes a `no` props to choose the slide title you want to render.
For example:
```vue
<script setup lang="ts">
import Titles from '/@slidev/titles.md'
<script>
<template>
  <Titles no="42"/>
  <Titles :no="item.path"/>
</template>
```
It's used by the the `<TocList>` component and I should be able to also use it into my external plugin.

For now I have a message:
```
Sourcemap for "/@slidev/titles.md" points to missing source files
```
I'll try to fix this.

tonai:
All fixed

tonai:
Sure

PR #489, opened 2022-02-18
fix: video having no duration (with CDN)

marob:

This PR imports a JS file from a CDN in `index.html`.
Cons: we depend on internet connection and CDN availability.

See other PRs:
- https://github.com/slidevjs/slidev/pull/490
- https://github.com/slidevjs/slidev/pull/491


antfu:

PR #490, opened 2022-02-18
fix: video having no duration (with recordrtc dependency from GitHub)

marob:

This PR imports `recordrtc` dependency twice (second time from GitHub URL instead of from npm registry) in order to have the required `EBML.js` file (which is not packed in the npm registry version of `recordrtc`)
Cons: we declare the same dependency twice and we use GitHub dependency (not a good practice?).

See other PRs:
- https://github.com/slidevjs/slidev/pull/489
- https://github.com/slidevjs/slidev/pull/491


antfu:

PR #491, opened 2022-02-18
fix: video having no duration (with ts-ebml dependency from GitHub)

marob:

This PR imports `ts-ebml` dependency from GitHub URL. It's using a fork of the original repository as it's using code from a [PR that is not merged yet](https://github.com/legokichi/ts-ebml/pull/38)

Cons: we use GitHub dependency (not a good practice?) and from a fork

See other PRs:
- https://github.com/slidevjs/slidev/pull/489
- https://github.com/slidevjs/slidev/pull/490


antfu:
If `ts-ebml` is no longer actively maintained, I am wonder if anyone would like to make a fork with the fix and publish it to npm. That way would be better to install from GitHub.

antfu:

PR #498, opened 2022-02-26
feat: support specify `list-class` for TOC

alexanderdavide:
Currently, using `<Toc />` merely offers a disc styled list but table of contents usually want numeric numbering.

With this PR, the `<Toc />` list style becomes configurable using the added prop `listStyle` which allows `decimal`and `disc` and defaults to `decimal`.

### DIFF
--- a/packages/client/builtin/Toc.vue
+++ b/packages/client/builtin/Toc.vue
@@ -15,14 +15,14 @@ import { tree } from '../logic/nav'
 const props = withDefaults(
   defineProps<{
     columns?: string | number
-    listClassNames?: string[]
+    listClass?: string | string[]
     maxDepth?: string | number
     minDepth?: string | number
     mode?: 'all' | 'onlyCurrentTree' | 'onlySiblings'
   }>(),
   {
     columns: 1,
-    listClassNames: () => [],
+    listClass: '',
     maxDepth: Infinity,
     minDepth: 1,
     mode: 'all',
@@ -80,6 +80,6 @@ const toc = computed(() => {
 
 <template>
   <div class="slidev-toc" :style="`column-count:${columns}`">
-    <TocList :level="1" :list="toc" :list-class-names="listClassNames" />
+    <TocList :level="1" :list="toc" :list-class="listClass" />
   </div>
 </template>
--- a/packages/client/builtin/Toc.vue
+++ b/packages/client/builtin/Toc.vue
@@ -15,12 +15,14 @@ import { tree } from '../logic/nav'
 const props = withDefaults(
   defineProps<{
     columns?: string | number
+    listStyle?: 'decimal' | 'disc'
     maxDepth?: string | number
     minDepth?: string | number
     mode?: 'all' | 'onlyCurrentTree' | 'onlySiblings'
   }>(),
   {
     columns: 1,
+    listStyle: 'decimal',
     maxDepth: Infinity,
     minDepth: 1,
     mode: 'all',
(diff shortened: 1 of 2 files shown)


alexanderdavide:
I though I'd make it simple, allowing only `disc`and `decimal` but you're probably right. What do you think of this version? `ol` by default and the following example would result in a disc list style handing over the classes the the `ol`.

```
<Toc :listClassNames="['!list-disc']" />
```

By the way, I am facing issues rebuilding the demo running `pnpm demo:dev`. Changing the slides markdown, saving is not enough for the changes to become visible. I am forced to give the local server a restart. Is this expected behavior?

twitwi:
I was passing around and saw this PR.
Maybe it is a useful helper but currently as vue propagates (to the root) the attributes passed to components, it can be done with the current implementation, e.g.:

```html
<Toc class="funky"/>
```

```css
.funky > ul {
    list-style: none;
}
.funky > ul ul {
    list-style: lower-greek;
}
```

twitwi:
For reference, similarly, pure CSS can be used to display the TOC as dots (see css below).

Compared to an experiment I had made with a TOC component https://twitwi.github.io/slidev-incubation/incubator/#/3 , I think it is not currently possible to style differently the previous and future items in the TOC (the `router-link-active` class is present on the `a` but the css cannot express "a li that is before a li that contains a a").

```css
.funky ul, .funky li {
  display: inline-block;
  font-size: 0;
}
.funky li a {
  display: inline-block;
  border-style: none !important; /* remove theme style */
  min-width: 10px;
}
.funky li a::after {
  content: '○';
  font-size: 10pt;
}
.funky li a.router-link-active::after {
  content: '◉';
}
.funky li li a::after {
  content: '◌';
}
.funky li li a.router-link-active::after {
  content: '◎';
}
.funky li a:hover::after {
  color: red;
}
```

alexanderdavide:
@twitwi Many thanks for joining the issue and bringing up these options.

My intent was to make changing the list style as easy as possible because I'd argue it is the most common aspect one would like to change of how `<Toc/>` is displayed. With this in mind, I would even consider my first approach again (27079ebefa1c249c6d392804f816c2e2cbb159f7) and add the rest of the list styles. Advanced styling of `<Toc />` could happen with CSS as @twitwi explains.

What do you all think?

twitwi:
(I had written a comment but I probably forgot to finish it... I'm trying to recreate it)

I have made a PR https://github.com/slidevjs/slidev/pull/502 mainly to render the TOC as a list of disks (without the text).
Currently it adds a mode but it might be better to make this rendering choice independently of the choice of what elements to show (i.e., the proposed mode="allDots" is currently exclusive with mode="onlyCurrentTree" etc).

Maybe this PR 502 can be cleaned and generalized to simplify many use cases, e.g. (it is always possible to use mode="onlyCurrentTree" ...).

```html
<Toc/>    the default, as now
```
Some "presets" for typical use

```html
<Toc render="disks"/>   also the default
<Toc render="dots"/>    the compact view as in 502
```
Customized things (used as in https://github.com/slidevjs/slidev/pull/502/files#diff-09ddd923cd380d377ae432ac8c9af51ee1854aebf653a8a79ab915da87101ef3R18 )

```html
<Toc render="I1o"/>           to have a full toc with roman numbers at level 1, arabic at level 2 and disks for all deeper levels
<Toc render="●○◌◡."/>     (same as render="dots", i.e. any non letter or digit character is used )
```

Issue #499, opened 2022-02-27
Table of content links keep/steal keyboard focus (mainly with firefox)


twitwi:
**Describe the bug**
Table of content links keep/steal keyboard focus.

**To Reproduce**
1. Open a presentation containing a `<Toc/>`
2. Click a link in the TOC (a `<RouterLink/>`)
3. Then try to use the keyboard to navigate the presentation

The keys do not navigate anymore in Firefox.
The same is true with chromium but only if the link is to the page itself (I guess that if the element becomes hidden, it is blur()'d in chromium, but not in firefox).

This is somewhat similar in behavior to https://github.com/slidevjs/slidev/issues/495

A solution is to add `@click="$event.target.blur()"` to the `<RouterLink/>` in the implementation of `<TocList/>`.
I'm not sure whether it is too hacky or not. If it is an acceptable solution, **I can PR that**.

**Desktop (please complete the following information):**
 - OS: Manjaro Linux
 - Browser: Firefox and Chromium (only when remaining on the same page)
 - Slidev version: 0.28.6

tonai:
Hello,
I tried to reproduce your problem but didn't find any issue.
I am able to navigate through the slides in Chrome or Firefox without any problem even if my presentation contains a `<toc>` or a `<RouterLink>` component

twitwi:
Hi,
Thanks for trying it.

Just to be sure there is no misunderstanding (I'm not sure from your message), the issue occurs when one click on the toc elements (step 2 of the reproduction instruction), before that the navigation works well.

Also, the issue is more problematic with Firefox; with chromium, it is problematic only in a degenerate case.

tonai:
Ok I understand.
I asked because I am also working to fix an issue (https://github.com/slidevjs/slidev/pull/513).
Just to inform you that in this PR I created a new Link component, and I updated the `<TocList>` component, so if both PR are approved, we will need to merge them.

twitwi:
I think your PR is very important. Fixing the focus (i.e. merging the two) is a very minor work, no worries.

tonai:
Can we close this issue ?

twitwi:
The issue is still there, but I'll (later) port the fix into the link component to force blur the router links. Thanks to your link component all this will be nicely factorized.

PR #500, opened 2022-02-28
fix: toc links keep keyboard focus in firefox

twitwi:
fixes https://github.com/slidevjs/slidev/issues/499

Here is a PR if it is ok to fix it this way (explicit inline $event.target.blur()).

Maybe something more general would be needed, e.g., I can imagine focusing the slide div when navigation occurs.


twitwi:
I've found it while implementing another feature related to the TOC that it could be better to not use RouterLink at all, and rather do a `<a @click="router.push({ path: item.path })">` because in some contexts `blur()` is not sufficient.
I'll make a PR later with the feature and a (coherent) fix replacing this one.

PR #502, opened 2022-03-01
feat: add toc compact mode

twitwi:
This PR adds a mode for toc `<Toc mode="allDots"/>` to show a concise TOC (e.g., to be put as a footer) that allows quick navigation and shows progress (inspired by beamer).

Possible design problems:
- there is some default style, maybe it should be in the theme?
- currently, the allDots is a mode so it is exclusive with other modes (that do not display all points)

This PR is adding some new links so it also fixes the focus problem #499, instead of adding more focus problem.

### DIFF
--- a/packages/client/builtin/Toc.vue
+++ b/packages/client/builtin/Toc.vue
@@ -17,7 +17,7 @@ const props = withDefaults(
     columns?: string | number
     maxDepth?: string | number
     minDepth?: string | number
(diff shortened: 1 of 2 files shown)


antfu:
Thanks for your contribution. But I'd prefer the more general #498 solution. That said, you can always copy the TOC component change as your want in userland.

Issue #483, opened 2022-02-15
PDF links to slides are broken


tonai:
**Describe the bug**
When creating a link in the slides with the `RouterLink` component (example `<RouterLink to="10">go to slide 10</RouterLink>`), the link does not work when exporting the slides into pdf. 

**To Reproduce**
1. Create a slidev project
2. Add `<RouterLink to="10">go to slide 10</RouterLink>` into one of the slide
3. It should work in dev and build mode (if you have at least 10 pages in your slides)
4. run `npm run export` and click the same link inside the PDF and it does not work

**Desktop (please complete the following information):**
- OS: ubuntu 20.04.3 LTS (Focal Fossa)
- Browser: chrome 96.0.4664.110
- Slidev version: main branch (v0.28.1)
- node: 16.13.1
- pnpm: 6.25.1

**Additionnal comments**
This is most likely due to the fact that the export creates each page independently and then concat them all to have the full pdf file.
Maybe we should consider having a custom `/print` URL that will format all the slides (using the `@media print`), so we can print the whole presentation using the browser print pdf feature.
It should also considerably increase the export time.

twitwi:
@tonai can this be closed?

tonai:
Sure

PR #513, opened 2022-03-09
feat: print all slides into one page, fix link in pdf export 

tonai:
This is currently a Work In Progress to try to solve the following issue https://github.com/slidevjs/slidev/issues/483

Currently this version adds a new route `/print` that displays all slides.

I created a new built-in `Link` component that uses anchor when we are in print mode.

When exporting to PDF, it uses the new route and the PDF is generated in one shot (no need to merge all pages together anymore).

Pros:
* Links works inside the PDF
* Big performance gain when generating the PDF (about 230 slides generated in 20 seconds against more than 11 minutes with the previous version)

Problems:
1. We need to rethink the nav logic because we can't rely on the global context to gives informations about `currentPage`...Etc.
   Currently the `currentPage` is 1 when printing, but this causes issue with the `<toc/>` and can cause issue with global components used in `global-top.vue` or `global-bottom.vue`.
   I propose we use `provide`/`inject` to give access to all those informations in child components.
   This way we can inject the correct values for each slides when we print them all together.
2. We need to duplicate some slides if we want to keep the `--with-clicks` print feature.
   It will need some work but I don't think this is an issue if we can use the previous proposal and send the `click` information with `provide`/`inject`.

@antfu What is your though about this ?

Thanks in advance for your answer.

### DIFF
--- a/packages/slidev/node/export.ts
+++ b/packages/slidev/node/export.ts
@@ -1,6 +1,6 @@
 import path from 'path'
 import fs from 'fs-extra'
-import { blue, cyan, green, yellow } from 'kolorist'
+import { blue, cyan, dim, green, yellow } from 'kolorist'
 import { Presets, SingleBar } from 'cli-progress'
 import { parseRangeString } from '@slidev/parser/core'
 import type { SlideInfo } from '@slidev/types'
@@ -33,7 +33,7 @@ function createSlidevProgress(indeterminate = false) {
   const progress = new SingleBar({
     clearOnComplete: true,
     hideCursor: true,
-    format: `  {spin} ${yellow('rendering')}${indeterminate ? '' : ' {bar} {value}/{total}'}`,
+    format: `  {spin} ${yellow('rendering')}${indeterminate ? dim(yellow('...')) : ' {bar} {value}/{total}'}`,
     linewrap: false,
     barsize: 30,
   }, Presets.shades_grey)
--- a/packages/slidev/node/export.ts
+++ b/packages/slidev/node/export.ts
@@ -1,6 +1,4 @@
 import path from 'path'
-import fs from 'fs-extra'
-import { PDFDocument } from 'pdf-lib'
 import { blue, cyan, green, yellow } from 'kolorist'
 import { Presets, SingleBar } from 'cli-progress'
 import { parseRangeString } from '@slidev/parser/core'
@@ -89,8 +87,9 @@ export async function exportSlides({
   const page = await context.newPage()
   const progress = createSlidevProgress()
 
-  async function go(no: number, clicks?: string) {
-    progress.update(no)
+  async function go(no: number | string, clicks?: string) {
+    if (typeof no === 'number')
+      progress.update(no)
 
     const path = `${no}?print${withClicks ? '=clicks' : ''}${clicks ? `&clicks=${clicks}` : ''}`
     const url = routerMode === 'hash'
@@ -156,40 +155,23 @@ export async function exportSlides({
   progress.start(pages.length)
 
   if (format === 'pdf') {
-    const buffers: Buffer[] = []
-    const genPdfBuffer = async(i: number, clicks?: string) => {
-      await go(i, clicks)
-      const pdf = await page.pdf({
-        width,
-        height,
-        margin: {
-          left: 0,
-          top: 0,
-          right: 0,
-          bottom: 0,
-        },
-        pageRanges: '1',
-        printBackground: true,
-        preferCSSPageSize: true,
-      })
-      buffers.push(pdf)
-    }
-    for (const i of pages)
-      await genPageWithClicks(genPdfBuffer, i)
-
-    const mergedPdf = await PDFDocument.create({})
-    for (const pdfBytes of buffers) {
-      const pdf = await PDFDocument.load(pdfBytes)
-      const copiedPages = await mergedPdf.copyPages(pdf, pdf.getPageIndices())
-      copiedPages.forEach((page) => {
-        mergedPdf.addPage(page)
-      })
-    }
-
-    const buffer = await mergedPdf.save()
     if (!output.endsWith('.pdf'))
       output = `${output}.pdf`
-    await fs.writeFile(output, buffer)
+
+    await go('print')
+    await page.pdf({
+      path: output,
+      width,
+      height,
+      margin: {
+        left: 0,
+        top: 0,
+        right: 0,
+        bottom: 0,
+      },
+      printBackground: true,
+      preferCSSPageSize: true,
+    })
   }
   else if (format === 'png') {
     await genPagePng(pages)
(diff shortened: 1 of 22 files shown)


tonai:
So I made a new commit to switch from the `$slidev` global property to a `provide` / `inject` context.
This way I am able to provide a context for each slides while I am rendering them all in the `/print` page.
To avoid breaking anything, I created a loader to automatically `inject` this context into all slides and all vue components so you can still use `$slidev` without worrying.
I updated the `<toc>` component to use this context instead of importing from `logic/nav`.
For now I did not update other components, but if we want to have them compatible with the print mode we should update them as well.
I will now check the `--with-clicks` feature.

tonai:
I added back the `--with-clicks` feature. :partying_face: 

I also fixed some issue I add with the `<Tweet>` and `<Youtube>` components.

And I switch the PNG export to use the same process (using `Locator.screenshot()` from Playwright).

Now export time is around 20-30s for both format (for slides in `demo/starter` folder).

What still needs to be done:
- probably add a `--with-drawing` CLI option to export the drawings (currently it is enabled by default)
- maybe we should refactor components to use `inject` instead of importing from `logic/nav`? I only refactored the `<Toc>` because components like the `<Editor>` does not make sense in print mode, but maybe we should follow the same logic for every components?
- I kept methods and computed values in `logic/nav` for retro-compatibility purpose but maybe we should remove them or mark them as deprecated?
- What to do with the progress bar displayed on the CLI while exporting ? We can't use the same logic now as everything is printed in one shot (we can use it for PNG as I am iterating over the slide items but not for PDF, or in an other way...).

Do you have any opinion @antfu ?

tonai:
Fixing merge conflicts

packages/slidev/node/export.ts
```diff
@@ -70,13 +69,17 @@ export async function exportSlides({
   timeout = 500,
   dark = false,
   routerMode = 'history',
-  width = 1920,
-  height = 1080,
+  width = 980,
+  height = 551,
```
antfu, line 73: What's the reason of this change?

packages/slidev/node/export.ts
```diff
@@ -89,39 +92,55 @@ export async function exportSlides({
...
     await page.goto(url, {
       waitUntil: 'networkidle',
     })
-    await page.waitForTimeout(timeout)
     await page.waitForLoadState('networkidle')
     await page.emulateMedia({ colorScheme: dark ? 'dark' : 'light', media: 'screen' })
+    // Wait for twitter iframe if we have some embedded tweets
+    const count = await page.locator('.tweet').count()
+    if (count > 0)
+      await page.locator('.tweet iframe').waitFor()
```
antfu, line 108: This is a bit too specific, possible to have a more general solution

packages/slidev/node/export.ts
```diff
@@ -89,39 +92,55 @@ export async function exportSlides({
...
-  function getClicks(url: string) {
-    return url.match(/clicks=([1-9][0-9]*)/)?.[1]
+  async function genPagePdf() {
+    if (!output.endsWith('.pdf'))
+      output = `${output}.pdf`
+    await go('print')
+    await page.pdf({
+      path: output,
+      width: 980,
+      height: 551,
```
antfu, line 122: Could we use the variable instead of hard-coded numbers?

antfu:
> What to do with the progress bar displayed on the CLI while exporting ? We can't use the same logic now that everything is printed in one shot (we can use it for PNG as I am iterating over the slide items but not for PDF, or in an other way...).

Possible to use a generator or event callback to notify the progress?

packages/slidev/node/export.ts
tonai: Because the slides size are 980/551 by default (from `canvasWidth` and `aspectRatio`).
There is no need to open a browser that is 1920 wide because we only use 980px width maximum.
I can change the numbers to use the config values instead.

tonai: So I can revert default values in fact, that's not a problem if you prefer so.

packages/slidev/node/export.ts
tonai: I would like but that's a fix that is specific for Twitter because of the use of the Twitter API.
The `<Tweet>` component does not render an iframe, it renders a `<div>` with an icon inside.
And without this, Playwright does not wait for the Tweeter component to completely render. (I see the carbon icon in the export)
So in fact I tell Playwright to wait for the twitter <iframe>` element to be present in the DOM (only if I use the `<Tweet>` component).
This does not happen with the `<Youtube>` since it already dislpays the `<iframe>` element.

The only way I think to make this more general it to use some data in the componen.
For example in the `<Tweet>` you can have something like:
```
<template>
  <Transform :scale="scale || 1">
    <div ref="tweet" class="tweet" data-waitfor="iframe">
      <div v-if="!loaded" class="w-30 h-30 my-10px bg-gray-400 bg-opacity-10 rounded-lg flex opacity-50">
        <div class="m-auto animate-pulse text-4xl">
          <carbon:logo-twitter />
        </div>
      </div>
    </div>
  </Transform>
</template>
```
So if we need to wait for something else we can reuse the `data-waitfor` attribute inside another component.
What do you think?

tonai:
> Possible to use a generator or event callback to notify the progress?

The problem is more with the steps.
As I render all the pages I an't use the slide number as the steps anymore.
So what steps should we use ?
I can use fixed steps:

* increment step after `await page.goto()`
* increment step after `await page.waitForLoadState()`
* increment step after `await page.emulateMedia()`
* increment step after `await page.locator('.tweet')`
* ...etc.

Or I can update the progress with an `indeterminate` option that will only displays the `{spin}` and `'rendering'` parts.

packages/slidev/node/export.ts
antfu: Could we generally wait for all iframes?

antfu:
> Or I can update the progress with an indeterminate option that will only displays the {spin} and 'rendering' parts.

Sounds good to me!

packages/slidev/node/export.ts
tonai: Waiting for all iframes is done on the line after:
```js
const frames = await page.frames()
```
Here we wait for the iframe to be present in the document, before waiting for the iframe to be loaded.
But anyway I will push a reusable solution, in case there is an other component that need to wait for specific DOM elements before printing the page.

tonai:
> Sounds good to me!

Good, I pushed all changes.

tonai:
> I'm receiving updates on this so I take the opportunity to add a request/question: would there be a way to not print some slides? is it just a question of adding a custom class like no-print and some CSS or should it be embedded in the print logic?

I think you can use `class: print:hidden` in the Front Matter block.

tonai:
> Thanks for the suggestion.
> I had a quick try (on slidev without this PR) and the slide is still printed in the pdf :(

I tested on this branch and it seems to work

twitwi:
Ok. I'll retry at some point. Sorry for polluting the discussion.

antfu:
Awesome work! @tonai Thanks for the contributions you have made to Slidev, I have sent you the invitation to join the org as a part of the team! 🙌

Issue #475, opened 2022-02-08
Add mp4 or mov exports to the recording


kissu:
When recording and sending the `.webm` to a conference after a record, they usually me for a `.mp4` or a `.mov` because it's probably a more common extension for video montage softwares.

Having the possibility to choose an extension before the recording would be perfect!

I could transcode it to another format but it will probably bring some quality loss during the conversion. It's still the only available solution right now, but the quality suffers out of this.

antfu:
AFAIK, most browsers only support recording to webm format, and I don't know how to convert them efficiently with JS

MarvinXu:
[RecordRTC](https://github.com/muaz-khan/RecordRTC) support sevaral Codecs and can be configured by `mimeType`.
So we can pass in a custome config into `packages\client\logic\recording.ts`

I'm not sure how to design this feature. Maybe adding a selection list in the recording dialog?

PR #519, opened 2022-03-19
feat: support mp4 when recording

MarvinXu:
feature demo #475
image

image


packages/client/internals/RecordingDialog.vue
```diff
@@ -18,15 +20,25 @@ const value = useVModel(props, 'modelValue', emit)
 
 const { startRecording } = recorder
 
+const mimeTypeItems = supportedMimeTypes.map(mime => ({
+  value: mime,
+  display: mime,
+}))
+
+const currentMimeType = ref('video/webm')
```
antfu, line 28: We could have it as a global ref with `useStorage` to persist the preference.

PR #521, opened 2022-03-21
feat: custom nav controls

LarchLiu:
I have a component that needs some buttons, and although I can put the buttons in the component, it only shows up in the slides wrapper and changes size as the wrapper resize. So I want to put these buttons into nav controls.


antfu:
Thank you. Can you also help providing some documents for it? Thanks!

LarchLiu:
Done it :)

Issue #530, opened 2022-04-02
Support for exporting speech transcripts from slidev


hangboss1761:
**Is your feature request related to a problem? Please describe.**
Support for exporting speech transcripts from slidev.

**Describe the solution you'd like**
Combine each slide with its notes to generate an `.md` file, use `slidev export --format md` to export, you can publish your speech transcripts as an article on any platform that supports `markdown`

the `.md` file like this
![image](https://user-images.githubusercontent.com/30379896/161384316-a9dc947f-a9ee-4ef7-b3b9-060fc07fee71.png)

PR #531, opened 2022-04-02
feat: slidev export support --format md (#530)

hangboss1761:


packages/slidev/node/export.ts
```diff
@@ -106,7 +109,39 @@ export async function exportSlides({
...
+    }
+    for (const i of pages)
+      await genPageWithClicks(genScreenshot, i)
+  }
+
+  async function genPageMd(pages: number[], slides: SlideInfo[]) {
+    const mds: string[] = []
+
+    for (const i of pages) {
+      const mdImg = `![](./${output}/${i.toString().padStart(2, '0')}.png)\n\n`
```
antfu, line 131: Do we also need the title of each slides here?

hangboss1761: title like `# Welcome to Slidev` ? 

It seems a bit redundant to force it

antfu: How about `${title}`? TBH I am not so confident if it could fits the general needs, as the current API there is no way to do customization.

hangboss1761: Yes, it looks better this way

hangboss1761: code updated

antfu (approved):

tonai:
Maybe add `slides-export.md` in the `.gitignore`.

Also you have incoherent result if you use the `--with-clicks` option.

hangboss1761:
> Maybe add `slides-export.md` in the `.gitignore`.
> 
> Also you have incoherent result if you use the `--with-clicks` option.

`slides-export.md` should be in the `.gitignore`

how about set `withClicks = false` when `format = md`

Issue #542, opened 2022-04-13
Timeout when exporting to PDF


tonai:
**Describe the bug**
Export can timeout if the number of slides is very important, or for small CPU (or both).

**To Reproduce**
1. Create a really big presentation 
2. Exports the slides to PDF using `slidev export` command
3. Try to limit the CPU usage of the node process (using for example `cpulimit`)
4. See the error:
```
page.goto: Timeout 30000ms exceeded.
=========================== logs ===========================
navigating to "http://localhost:12445/print?print", waiting until "networkidle"
============================================================
```

**Desktop (please complete the following information):**
 - OS: Ubuntu 20.04
 - Slidev version: 0.30.0

Issue #116, opened 2021-05-16
can not export to pdf


ahangchen:
I try to use `npx slidev export` to export pdf, but it always fails on:

```
 ▲ rendering █████░░░░░░░░░░░░░░░░░░░░░░░░░ 1/6page.goto: Timeout 30000ms exceeded.
=========================== logs ===========================
navigating to "http://localhost:12445/1?print", waiting until "networkidle"
============================================================
Note: use DEBUG=pw:api environment variable to capture Playwright logs.
Error
    at Object.captureStackTrace (/Users/cwh/Mission/coding/slidev/node_modules/playwrigh)
    at Connection.sendMessageToServer (/Users/cwh/Mission/coding/slidev/node_modules/pla)
    at Proxy.<anonymous> (/Users/cwh/Mission/coding/slidev/node_modules/playwright-chrom)
    at /Users/cwh/Mission/coding/slidev/node_modules/playwright-chromium/lib/client/fram5
    at Frame._wrapApiCall (/Users/cwh/Mission/coding/slidev/node_modules/playwright-chro)
    at Frame.goto (/Users/cwh/Mission/coding/slidev/node_modules/playwright-chromium/lib)
    at /Users/cwh/Mission/coding/slidev/node_modules/playwright-chromium/lib/client/page0
    at Page._attributeToPage (/Users/cwh/Mission/coding/slidev/node_modules/playwright-c)
    at Page.goto (/Users/cwh/Mission/coding/slidev/node_modules/playwright-chromium/lib/)
    at go (/Users/cwh/Mission/coding/slidev/node_modules/@slidev/cli/dist/export-WGBEP3O)
    at exportSlides (/Users/cwh/Mission/coding/slidev/node_modules/@slidev/cli/dist/expo)
    at async Object.handler (/Users/cwh/Mission/coding/slidev/node_modules/@slidev/cli/d{
  name: 'TimeoutError'
```

antfu:
Can't reproduce, what's the content of your slides, does it do with any web requests?

LeadcodeDev:
I have the same error with the default install :
• yarn create slidev (don't run install dependencies & start)
• yarn install
• yarn export

cf : https://fr.sli.dev/guide/#creation-de-votre-premiere-presentation

ahangchen:
May be caused by unreachable internet connetion, when rendering with some plugins. 
I turn on a global proxy and the progressbar moves.

mariomurrent-softwaresolutions:
[...]

PR #543, opened 2022-04-13
fix: use timeout arg for the goto() function

tonai:
Fixes https://github.com/slidevjs/slidev/issues/542

With this PR https://github.com/slidevjs/slidev/pull/513, the `await page.waitForTimeout(timeout)` does not seems to be needed anymore.
The goal of this PR is to reuse the `timeout` arg you can pass to the `slidev export` command to be able to increase the timeout of `page.goto()` and to be able to generate the export in failing scenarios.

### DIFF
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -303,7 +303,7 @@ cli.command(
       describe: 'output format',
     })
     .option('timeout', {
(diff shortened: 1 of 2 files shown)


tonai:
Should also fix https://github.com/slidevjs/slidev/issues/116

PR #550, opened 2022-04-14
feat: add option to allow recording in static build

### LINKED ISSUE #375: Allow to record on production
Hello,

I couldn't find in the docs if there is a way to allow the user to use the record feature if the presentation is build as SPA and deployed to a hosting.

If that isn't possible I think would be great to have something similar to the `download` feature so the user can turn on/off this feature on the frontmatter info

tonai:
fixes #375


tonai:
Doc updated https://github.com/slidevjs/docs/pull/65

Issue #24, opened 2021-05-07
Presenter mode isn't working afrer build?


xorik:
Hi!

First of all: thank you for the great project! I want to make a short presentation on my work about a big refactor, and I always hated making presentations. But now I really enjoy, especially the line highlight feature.

Steps to reproduce:

```shell
npm run build
```
put file `dist/serve.json`

```
{
  "rewrites": [{
    "source": "**",
    "destination": "index.html"
  }]
}
```

```shell
cd dist
npx serve
```

Go to http://localhost:5000/, make sure routing is working after page reload
Go to http://localhost:5000/presenter - it redirects to the first slide.

I know that I can start a dev server for presenter mode, but it's over 5 megabytes without images. Looks like there is no compression?

twitwi:
maybe, using the inter-window messaging API would allow for a serverless solution https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage (which is what is done in some other HTML slide frameworks)

ennioVisco:
The new multi-screen api should fit the need for enabling it in only client-side scenarios: https://web.dev/multi-screen-window-placement/

PR #552, opened 2022-04-15
Feat/standalone presenter mode

tonai:
Fixes https://github.com/slidevjs/slidev/issues/24

The idea is to use different technologies depending on the cases:
* dev mode: continue to use server state with`vite-plugin-vue-server-ref` to keep the possibility to remotely control the slides using the `--remote` option
* build mode with persist drawing option enabled: use `localStorage` to synchronize the state between multiple tabs
* build mode with persist drawing option disabled: use `BroadcastChannel` to synchronize the state between multiple tabs

For both `localStorage` and `BroadcastChannel`, I use a key based on the slides main title instead of using a `slidev-` prefixed key.
This prevent to have different slidev presentation opened that interact which each others.

### DIFF
--- a/packages/client/env.ts
+++ b/packages/client/env.ts
@@ -3,12 +3,8 @@ import { computed } from 'vue'
 import { objectMap } from '@antfu/utils'
 // @ts-expect-error missing types
 import _configs from '/@slidev/configs'
-import _serverDrawingState from 'server-reactive:drawings?diff'
-import type { ServerReactive } from 'vite-plugin-vue-server-ref'
 
-export const serverDrawingState = _serverDrawingState as ServerReactive<Record<number, string | undefined>>
 export const configs = _configs as SlidevConfig
-
 export const slideAspect = configs.aspectRatio ?? (16 / 9)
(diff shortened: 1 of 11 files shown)


antfu:
My concern would be in the build mode it won't work with multiple devices, and creates a misalignment to the dev mode. I am thinking to have some provider interface allowing customize the services providers like Firebase or Supabase to do the online syncing. In addition to that, we might need a password for the presenter mode in production.

tonai:
Hello,
I am back from some vacation :)

I understand your point, but this PR is still an enhancement that fix https://github.com/slidevjs/slidev/issues/24
If you want to work with multiple devices, you will need a server at some point.
And I think some people just want to have presenter mode working in local build, without the need of working with multiple devices (as I do).

For now I can understand as a user that I need to use the dev mode to be able to work with multiple devices.
That's an advanced feature that requires a server after all.
But presenter mode is a more basic feature that is mandatory for a presentation tool.

So maybe we can still merge this, and think to add the multiple device feature in build mode later ?
What do you think ?

PR #556, opened 2022-04-22
fix(demo): Compatible with prettier

Fiv5:
see #202


antfu:
My main concern about this would be it could mislead the user the the string after `---` is requires and has some sort of effects (as title or something)

antfu:
Thanks for contribution. Close due to inactive.

Issue #422, opened 2021-11-19
how to forbid others presenter mode?


yuyuexi:
Hi, there. Thanks for this great work and I really enjoy this project!

I am not sure if this is a feature or I can find someway to fix it: I want to share my slides to other audiences, but it is annoying that everyone accessible to this slides can use the presenter mode. Is there any method I can forbid this and only use the presenter mode by myself?

antfu:
I guess we could have something like http://localhost:3000/1?view that hides the UI so you can share the URL to others? If so, I am open for PR :)

yuyuexi:
> I guess we could have something like http://localhost:3000/1?view that hides the UI so you can share the URL to others? If so, I am open for PR :)

Thanks for your timely reply!

I tried [](http://localhost:3000/1?view), however I can also enter presenter mode from it. I do not know if I understand your suggestion correctly?

antfu:
Yes, it's not there for now - I was proposing a feature. If it looks good to you, maybe you can try to implement it and send a PR.

tonai:
I think this is an interesting issue when you want to give access to your presentation to others, but don't want them to go to presenter view and start messing the main presentation you are doing.
I don't think having a parameter in the URL is enough, because someone can still change it and go to the presenter view.

Maybe we can have a workflow like this one:
* add a frontmatter option in the slides like `privacy: 'private'` (will be `'public'` by default)
* if privacy is `'private'` then:
  * hide the presenter button in the control bar
  * generate a token and display it when launching the dev server: `presenter mode  > http://localhost:3030/presenter?token=abc`
  * only allow access to `/presenter` to people who have the right token in the URL (only if you are on dev mode)

What do you think ?

PR #576, opened 2022-05-09
feat: private presenter mode

tonai:
fixes https://github.com/slidevjs/slidev/issues/422

Add option to set presenter mode private in front matter.
In that case the app generates a token you should add to the URL to be able to access the presenter mode.
The token is displayed on the log when you start the server:
Capture d’écran de 2022-05-09 14-54-26

Of course this is more useful when you start the server with the `remote` option.
Note that this is not a full secure solution as the test is done on the frontend side, so someone can still hook into the the JS and access the presenter mode, but maybe this is enough ?

But having the test on the frontend side gives the ability to enable this option in the static build.
In that case the token is also given when building the app:
Capture d’écran de 2022-05-09 14-54-06

### DIFF
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -53,8 +53,7 @@ cli.command(
       describe: 'open in browser',
     })
     .option('remote', {
-      default: false,
-      type: 'boolean',
+      type: 'string',
       describe: 'listen public host and enable remote control',
     })
     .option('log', {
@@ -94,7 +93,7 @@ cli.command(
     async function initServer() {
       if (server)
         await server.close()
-      const options = await resolveOptions({ entry, theme }, 'dev')
+      const options = await resolveOptions({ entry, remote, theme }, 'dev')
       port = userPort || await findFreePort(3030)
       server = (await createServer(
         options,
@@ -103,7 +102,7 @@ cli.command(
             port,
             strictPort: true,
             open,
-            host: remote ? '0.0.0.0' : 'localhost',
+            host: remote !== undefined ? '0.0.0.0' : 'localhost',
             force,
           },
           logLevel: log as LogLevel,
@@ -389,8 +388,7 @@ function commonOptions(args: Argv<{}>) {
     })
 }
 
-function printInfo(options: ResolvedSlidevOptions, port?: number, remote?: string | boolean) {
-  const query = options.data.config.presenter.isPrivate ? `?token=${options.data.config.presenter.token}` : ''
+function printInfo(options: ResolvedSlidevOptions, port?: number, remote?: string) {
   console.log()
   console.log()
   console.log(`  ${cyan('●') + blue('■') + yellow('▲')}`)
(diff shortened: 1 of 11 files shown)


packages/types/src/config.ts
```diff
@@ -136,6 +136,12 @@ export interface SlidevConfig {
    * @default 'dev'
    */
   record: boolean | 'dev' | 'build'
+  /**
+   * Options for presenter mode
+   *
+   * @default {public:true}
```
antfu, line 142: The default does not match with types I think

tonai: Thanks for noticing, I forgot to update here

packages/types/src/config.ts
```diff
@@ -217,6 +223,21 @@ export interface DrawingsOptions {
   syncAll?: boolean
 }
 
+export interface PresenterOptions {
+  /**
+   * Restrict access to presenter mode
+   *
+   * @default true
+   */
+  isPrivate?: boolean
```
antfu, line 232: Maybe just `private`? And I would prefer to have it default to `false` as you already need to explicitly pass `--remote` flag.

Or more over, maybe we could unify this with `token` to a single entity called `password`. And we could have the cli to be `--remote mypassowrd`.

By that, we could have a basic auth header to prompt for passing the password when visiting the route. (Or just have it in the UI)

tonai: It is false by `default` in fact, I just forgot to update the default value in that (it has been fixed in the last commit).
I agree that this feature is only useful when having the `--remote` option, so changing to `--password` makes sense.
In that case I will remove this feature on the static build part until we have a solution to make the `--remote` option work..
I personally like the fact of having the token/password in the URL. This way you can just start the server, click on the link and avoid the additional step of typing the password.
But yes maybe we can add a prompt to enter the password in case it is not present in the URL.

antfu: We could go both ways. Could read from the URL, and when it's not presented we show and UI for input.

packages/types/src/config.ts
```diff
@@ -217,6 +223,21 @@ export interface DrawingsOptions {
   syncAll?: boolean
 }
 
+export interface PresenterOptions {
```
antfu, line 226: We could remove this then?

packages/types/src/config.ts
```diff
@@ -136,6 +136,12 @@ export interface SlidevConfig {
    * @default 'dev'
    */
   record: boolean | 'dev' | 'build'
+  /**
+   * Remote password
+   *
+   * @default undefined
+   */
+  remote?: string
```
antfu, line 144:
```suggestion
  /**
   * Expose the server to inbound requests (listen to `0.0.0.0`)
   *
   * Pass a string to set the password for accessing presenter mode.
   *
   * @default false
   */
  remote?: string
```

packages/types/src/config.ts
tonai: oops

packages/slidev/node/cli.ts
```diff
@@ -53,8 +53,7 @@ cli.command(
       describe: 'open in browser',
     })
     .option('remote', {
-      default: false,
-      type: 'boolean',
+      type: 'string',
```
antfu, line 56: Can this be an optional string?

tonai: Yes.
The only option I found for that is to use `'string'` without any default value:
* if no option => `remote=undefined` => remote not enabled
* if `--remote` => `remote=''` (empty string) => remote enabled and public
* if `--remote password` => `remote='password'` => remote enabled with password

PR #601, opened 2022-05-23
feat(mermaid): default dark theme, close #599

### LINKED ISSUE #599: Why switching themes doesn't take effect for mermaid?
as the title

I originally considered using media queries to dynamically modify svg, but because of my low CSS level and too many colors involved, I had to give up. Later, after researching, I found that the component inner-html was used. Just re-render the component with parameters.

But I tried to pass the mermaid native parameter when the component is built after pulling all the code, but no matter whether I pass 'darkMode: true' or directly modify 'background', it has no effect.

```ts
const svgObj = computed(() => renderMermaid(props.code || '', Object.assign({ theme: props.theme, background: '#000', darkMode: true }, vm!.attrs)))
const svgObj1 = computed(() => renderMermaid(props.code || '', Object.assign({ theme: props.theme, background: '#f4f4f4', darkMode: false }, vm!.attrs)))
console.log(svgObj.value===svgObj1.value)
const html = computed(() => svgObj.value)
const html1 = computed(() => svgObj1.value)
....
<ShadowRoot class="mermaid" :inner-html="html" />
<ShadowRoot class="mermaid" :inner-html="html1" />
....
```
image

console logs false for sure, but the image display should be different anyway, right?

Also, the slidev version I am using is 0.32.1 and the mermaid version is 9.1.1, both are latest version.

I thought I could mention a pr after the implementation, but it is invalid to pass native parameters directly, which makes me unable to deal with it.

htharoldht:
Change mermaid theme when switching between light and dark

### DIFF
--- a/packages/client/builtin/Mermaid.vue
+++ b/packages/client/builtin/Mermaid.vue
@@ -26,7 +26,13 @@ const props = defineProps<{
 
 const vm = getCurrentInstance()
 const el = ref<HTMLDivElement>()
-const svgObj = computed(() => renderMermaid(props.code || '', Object.assign(props.theme ? { theme: props.theme } : { theme: (isDark.value ? 'dark' : props.theme) }, vm!.attrs)))
+const svgObj = computed(() => renderMermaid(
+  props.code || '',
+  {
+    theme: props.theme || (isDark.value ? 'dark' : undefined),
+    ...vm!.attrs,
+  },
+))
 const html = computed(() => svgObj.value)
 const actualHeight = ref<number>()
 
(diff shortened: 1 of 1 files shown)


packages/client/builtin/Mermaid.vue
```diff
@@ -25,7 +26,7 @@ const props = defineProps<{
 
 const vm = getCurrentInstance()
 const el = ref<HTMLDivElement>()
-const svgObj = computed(() => renderMermaid(props.code || '', Object.assign({ theme: props.theme }, vm!.attrs)))
+const svgObj = computed(() => renderMermaid(props.code || '', Object.assign({ theme: (isDark.value ? 'dark' : props.theme) }, vm!.attrs)))
```
antfu, line 29: The user-specified theme should take the highest priority. We should change the theme as default.

Issue #625, opened 2022-06-09
[mermaid] scaling is broken


TheAlexLichter:
**Describe the bug**
The `scale` property is ignored since #570 has been in place. I guess because the SVG is now in the shadow root, height is not applied "correctly"?

**To Reproduce**

1. Check https://stackblitz.com/edit/slidev-xf3l8y?file=package.json,slides.md - it works fine 
2. Check https://stackblitz.com/edit/slidev-z9jj6c?file=package.json,slides.md,README.md where scaling is broken

cc @twitwi - any idea how to solve this in a elegant manner?

twitwi:
I'll have a look.

twitwi:
@manniL if it is easy for you to check that the branch solves the issue on your side, that would be amazing (it works with my local tests).

twitwi:
(unfortunately it might not be so easy to actually depend on the branch without rebuilding it manually)

manniL:
Yes, this fixes the issue! Thanks a lot 🙏 

(I monkeypatched the code to try out 🙈)

twitwi:
Amazing :) . Thank you.

PR #626, opened 2022-06-09
[fix] forgot to handle mermaid scale parameter in ShadowRoot

twitwi:
fixes https://github.com/slidevjs/slidev/issues/625

The scaling code was getting ignored due to a missing ref (the svg used to be ref="el" but was moved inside the shadow root).

### DIFF
--- a/packages/client/builtin/Mermaid.vue
+++ b/packages/client/builtin/Mermaid.vue
@@ -59,5 +59,5 @@ watchEffect(() => {
 </script>
 
 <template>
-  <ShadowRoot class="mermaid" :inner-html="html" />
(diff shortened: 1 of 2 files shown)


twitwi:
From what I know, there are two hacky solutions:
- monkey-patching the distribution (as per the comment on #625 )
- use a custom mermaid component that includes the patch (putting the [attached files](https://github.com/slidevjs/slidev/files/8912046/mymermaid.zip)
 in your components/ folder and using a component instead of a code block)

```xml
<MyMermaid :scale="0.2" code="graph TD
B[Text] --> C{Decision}
C -->|One| D[Result 1]
C -->|Two| E[Result 2]
"/>
```

EkaanshArora:
monkey-patching worked, had to clean the vite cache, thanks again :)

PR #659, opened 2022-07-19
feat: override shortcuts

### LINKED ISSUE #555: [Custom Shortcuts] `nav.toggleOverview()` does only toggle once
**Describe the bug**

When mapping a key to use `nav.toggleOverview()`, the overview is only toggled once.

Shortcut file:
```ts
import { defineShortcutsSetup, NavOperations } from '@slidev/types'

export default defineShortcutsSetup((nav: NavOperations) => {
  return [
    {
      key: 'Escape',
      fn: () => nav.toggleOverview(),
    }
  ]
})
```

**To Reproduce**
1. Create a new slidev project
2. Setup your `setup/shortcuts.ts` file as shown above.
3. Start slidev
4. Push ESC. You see the open overview now.
5. Push ESC again. It is still open.

 - OS: macOS
 - Browser: Chrome and FF, both show the issue
 - Slidev version: v.0.30.2

manniL:
Resolves #555

This PR alters the shortcut setup so that users can change the default shortcuts via `defineShortcutsSetup`, e.g. removing some or disabling all of them.


packages/client/logic/shortcuts.ts
```diff
@@ -49,29 +49,31 @@ export function registerShortcuts() {
...
-      { key: and(g, not(drawingEnabled)), fn: () => showGotoDialog.value = !showGotoDialog.value },
-      { key: and(left, showOverview), fn: prevOverviewPage },
-      { key: and(right, showOverview), fn: nextOverviewPage },
-      { key: and(up, showOverview), fn: upOverviewPage },
-      { key: and(down, showOverview), fn: downOverviewPage },
-      { key: and(enter, showOverview), fn: () => { go(currentOverviewPage.value); showOverview.value = false } },
-      ...customShortcuts,
-    ]
-      .map((options: ShortcutOptions) => [options.key, options]),
+    __SLIDEV_FEATURE_OVERRIDE_DEFAULT_SHORTCUTS__
```
antfu, line 52: This may not need to be a compile-time flag (it was mainly for tree-shaking necessary packages).

You can read it from `import { configs } from '../env'`

packages/types/src/config.ts
```diff
@@ -129,10 +129,15 @@ export interface SlidevConfig {
...
+  /**
+   * Options for shortcuts
+   * @default {}
+   */
+  shortcuts: ResolvedShortcutOptions
   /**
    * Options for drawings
    *
-   * @default {}
+   * @default { override: false }
```
antfu, line 140: 👀 wrong place?

packages/types/src/config.ts
```diff
@@ -129,10 +129,15 @@ export interface SlidevConfig {
    * @default 'https://cdn.jsdelivr.net/gh/slidevjs/slidev/assets/favicon.png'
    */
   favicon: string
+  /**
+   * Options for shortcuts
+   * @default {}
+   */
+  shortcuts: ResolvedShortcutOptions
```
antfu, line 136: While I do agree to make it an option would help on future extensive, this option might mislead users that you may config shortcuts in the frontmatter.

I guess maybe, instead, we could pass the default shortcut array as an argument to `setupShortcuts` so the user could mutate it or disable completely setting `arr.length = 0`. Wdyt?

packages/types/src/config.ts
manniL: Whoops, yes 🙈

packages/types/src/config.ts
manniL: Uhh, that's even better! Great idea 👍 
Will refactor ☺️

manniL:
Okay, I've changed the implementation and passed the default shortcuts to the `setupShortcuts` function accordingly so users can manipulate them.

Because injections only allow a single argument though, I had to change the signature of a the method which would lead to a breaking change.

antfu (approved):

Issue #662, opened 2022-07-26
feature request: Export PDF to generate TOC(outline, bookmarks)


condorheroblog:
**Problems encountered**

When there are many PDF pages, I want to quickly locate the specified page, because there is no directory and I need to scroll many times to find the page I need

**Describe the solution you'd like**

Choose to generate a Table of Contents (or bookmark) when exporting PDF from Slidev. can solve the problem of finding too many pages

<img width="1125" alt="image" src="https://user-images.githubusercontent.com/47056890/180923641-968975e3-d15f-4641-acfb-01543b2745cf.png">

Shinyaigeek:
https://github.com/puppeteer/puppeteer/issues/1778#issuecomment-1198240981 This comment in the issue of puppeteer may help. I try to work on this.

PR #669, opened 2022-07-31
feat(export): export pdf with outlines if `--with-toc` option is specified

Shinyaigeek:
## Motivation 🔥 

Fixes: #662 

I implemented a feature to export pdf with outlines. A user can generate a pdf with outlines if they specify `--with-toc` option in command line.

I happened to see the issue, and it seemed interesting, so I started working on it, but it went better than I expected, so I'm submitting it as a PR. I would like to discuss the implementation policy using this branch as a starting point. Please feel free to comment on any concerns or overlooked points.


packages/slidev/node/export.ts
```diff
@@ -133,6 +148,18 @@ export async function exportSlides({
       printBackground: true,
       preferCSSPageSize: true,
     })
+
+    if (withTOC) {
+      const outlinePdf = outlinePdfFactory(pdfLib)
+      const pdf = fs.readFileSync(output, { encoding: 'base64' })
+      const outline = slides.map(({ title }, idx) =>
+        title ? `${idx + 1}||${title}` : null,
+      ).filter(outline => !!outline).join('\n')
```
tonai, line 157: Don't you want to manage the "deepness" of title ?
In that case you may also be interested in the `level` property.

Also please heck the following properties (`rawTree`, `treeWithActiveStatuses` and `tree`) from here https://github.com/slidevjs/slidev/blob/main/packages/client/logic/nav.ts#L79-L86
I think It would better to use the same logic and thus the same functions than the ones used for the `<Toc>` component
And then having a dedicated renderer for the outline.

condorheroblog:
I want to say that `@lillallol/outline-pdf` is not perfect. The outline generated by `@lillallol/outline-pdf` can only jump to the corresponding page, but in the pdf editor, a good outline can jump to the corresponding part of the corresponding page.

image

packages/slidev/node/export.ts
Shinyaigeek: Thank you for your comment! I did not know `<Toc />` component. I will refer it and change my logic to use same logic of `<Toc />`. However, I am wondering how to use `rawTree` in nodejs `slidev` because `rawTree` is plugged via a vite slidev plugin in client side build. 💭

Shinyaigeek: @tonai 

I'm sorry for the late response 🙇 

I changed to handle "deepness" of title in [18cf377](https://github.com/slidevjs/slidev/pull/669/commits/18cf377ea43a853503d8eb8be7a233fd829fabea).
スクリーンショット 2022-08-14 17 52 37

> Also please heck the following properties (rawTree, treeWithActiveStatuses and tree) from here https://github.com/slidevjs/slidev/blob/main/packages/client/logic/nav.ts#L79-L86
I think It would better to use the same logic and thus the same functions than the ones used for the <Toc> component
And then having a dedicated renderer for the outline.

I agree, but the current logic about `<Toc />` is too for client-side to share the logic also with `export` command in CLI…

To share the logic about `<Toc />` also with `export` command, I searched a package to share the logic in client-side and server-side, but I could not find that… Do you know sucha a package for this usecase?

Shinyaigeek: FYI: in the current implementation of this PR, the logic about ToC of PDF in `export` command is implemented adhocky separated from `<Toc />` component.

tonai: @Shinyaigeek The idea of reuseing the same functions would be to ensure that the generated TOC will be the same as the one created with the `<Toc/>` component.
But yes those functions live in the client package.

@antfu what do you think of having a shared module that could be imported in either the backend or the frontend ?
Or should we duplicate some functions ? (here it is about all 3 `addToTree`, `getTreeWithActiveStatuses` and `filterTree` functions in `client/logic/nav.ts`)

Shinyaigeek: Yes, as @tonai says, it is better to share the logic which is reusable. If there is no package to share the utility with client-side and cli, perhaps we should make such a package before promoting this PR.

Shinyaigeek:
Not stale

antfu:
@tonai feel free to merge when this looks good to you

antfu (approved):

PR #677, opened 2022-08-07
feat: add chrome-executable parameter to export

fabianhauser:
This adds a `export --chrome-executable` CLI parameter to override the bundled chrome coming with playwright.

I use this to export my slides with the chrome provided by my system, as the playwright chrome build is not binary compatible (I use nix for building, but this would also apply for a musl based system like alpine linux.)

### DIFF
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -322,6 +322,10 @@ cli.command(
       type: 'boolean',
       describe: 'export pages for every clicks',
     })
+    .option('executable-path', {
+      type: 'string',
+      describe: 'executable to override playwright bundled browser',
+    })
     .strict()
     .help(),
   async ({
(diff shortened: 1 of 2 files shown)


antfu (approved):

packages/slidev/node/cli.ts
```diff
@@ -322,6 +322,10 @@ cli.command(
       type: 'boolean',
       describe: 'export pages for every clicks',
     })
+    .option('chrome-executable', {
```
antfu, line 325: Maybe `executable-path` to match with playwright?

packages/slidev/node/cli.ts
```diff
@@ -322,6 +322,10 @@ cli.command(
       type: 'boolean',
       describe: 'export pages for every clicks',
     })
+    .option('executable-path', {
+      type: 'string',
+      describe: 'executable to override playwright bundled chrome',
```
antfu, line 327:
```suggestion
      describe: 'executable to override playwright bundled browser',
```

fabianhauser:
Done, thanks for the review!

Issue #629, opened 2022-06-14
Allow both custom and addons/themes keyboard shorcuts


twitwi:
**Is your feature request related to a problem? Please describe.**
When an addon (or probably theme) defines keyboard shortcuts and the presentation author also wants to define custom shortcuts, the addon shorcuts are ignored.
Similarly (probably), only one addon can contribute keyboard shorcuts.

**Describe the solution you'd like**
It would be nice if the shortcuts were merged.
And ideally, when a key is bound both in the addon and presentation repo, the later should be used.

**Describe alternatives you've considered**
It is always possible to (manually) "eject" and merge the addons shortcut.

PR #702, opened 2022-09-03
feat!: injection chains to allow merging custom and addons/themes keyboard shortcuts

twitwi:
Fixes https://github.com/slidevjs/slidev/issues/629

IMPORTANT, There are 4 commits
- the 1st one is the core feature, probably without big issues,
- the 2nd one is an (optional but that I find nice) refactoring,
- the 3rd one is **to be discussed** (or just dropped) as it is a **breaking change** and would necessarily require a documentation update (that would be highly recommended anyways even with the first commits).
- following the discussion commit 4 adds a user message to help migration

Basically, the third commit breaks every situation that used setup/shortcuts (the default shortcuts are then removed).
The doc update would suggest something like:
```ts
export default defineShortcutsSetup((nav: NavOperations, baseShortcuts: any) => {
  return [
    ...baseShortcuts,
    {
```

Related to https://github.com/slidevjs/slidev/pull/659 (authored by @manniL )
- I kept the multi-args logic even if (with the 3rd commit) it is not used anymore, because it can become useful
- to remove some shortcuts, the principle would be to remove from baseShortcuts before the return with ...baseShortcuts (i.e. there is no "global" object that we mutate, only the return value is used)

### DIFF
--- a/packages/client/logic/shortcuts.ts
+++ b/packages/client/logic/shortcuts.ts
@@ -42,10 +42,10 @@ export function strokeShortcut(key: KeyFilter, fn: Fn) {
 }
 
 export function registerShortcuts() {
-  const { customShortcuts, defaultShortcuts } = setupShortcuts()
+  const allShortcuts = setupShortcuts()
 
   const shortcuts = new Map<string | Ref<Boolean>, ShortcutOptions>(
-    [...defaultShortcuts, ...customShortcuts].map((options: ShortcutOptions) => [options.key, options]),
+    allShortcuts.map((options: ShortcutOptions) => [options.key, options]),
   )
 
   shortcuts.forEach((options) => {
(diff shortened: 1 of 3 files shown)


twitwi:
Following the failing cypress, I'll force push the replacement of `replaceAll` by `replace`.

antfu:
I am not sure if the breaking change is necessary, but @manniL since you are using the feature, I guess we could have it if it make sense to you

manniL:
I'm fine with the suggested changes as they put defining shortcuts a little bit more "in line" with other features, especially regarding "implicit" mutation of the shortcuts object ☺️

twitwi:
Another solution, which is better for backward compatibility and at the same time allow the new API... but is maybe over-engineered given than slidev is still in version 0.*... would be to keep the old API untouched and rename this one as setupShortcuts2 or something like that.

This could be combined with runtime messages and progressive deprecation... but again it is probably too much here.

The solution I'd (now) recommend is to do the breaking change + add a check: if all shortcuts have been removed (not returned by the injected function) then we warn the user (how?) with a link to the updated doc.

twitwi:
@antfu considering the discussion, I'd say this one is ok to merge together with the doc PR mentioned). As it is a breaking change, I leave you the final click.

Issue #700, opened 2022-09-02
Allow (pre)parser extension


twitwi:
**Is your feature request related to a problem? Please describe.**
For some presentation that aggregate subslides, it takes a lot of space to even write the front matter with the two `---` (one after the frontmatter, and one after the (empty) content) + the empty line.
There are other cases where I'd like to have compact notations.

**Describe the solution you'd like**
A generic, minimal change solution, would be to add an extension point to the parser, that addon and user can customize.
E.g. allow to call a (chain of) custom parsers for instance here https://github.com/slidevjs/slidev/blob/fb8054f376b3b71c594b349fa7339cf2e2344a7e/packages/parser/src/core.ts#L130

**Describe alternatives you've considered**
I could write a script (bash, python) that generates the actual md file...

antfu:
Have you tried to write an inline Vite plugin to do the transformation? If that can't help, I am fine to expose some capability for such customization

twitwi:
I'll have a look at what can be done with vite, it is still some tech I don't know very well.

twitwi:
@antfu can vite plugins "intercept" a file read with `fs` as done in this case? https://github.com/slidevjs/slidev/blob/df066d434468160795e5718d22d08d450e2dd1a9/packages/parser/src/fs.ts#L9

I tried a simple plugin but it seems to be called with individual virtual 1.md 2.md ... files (but not the entry point).

antfu:
I see. PR welcome to propose your interface in mind (or if you want to start with a more detailed API design first would also be great)

twitwi:
(I use the term parser here)

I started thinking about it, I imagine a few callbacks from the parser to the plugin, do you prefer a style where there is an api with several methods or rather a single method with an kind of event type (i.e. the plugin implements a single method, that is a if (type == '') .... elif ....)?

For the "chain" aspect, where e.g. several (slidev) addons can each add a parser plugins and the user folder can also add parser plugins, I think the core of the solution is related to the same question with shortcuts https://github.com/slidevjs/slidev/issues/629 ... any suggestions on how you would do that best?

twitwi:
NB: the situation is even more tricky as:
- to know which addons are used we need the headmatter (frontmatter of the first slide)
- so the parser basically should run addon-less until it has read the headmatter,
- then only it can load the addon, asynchronously, and using config information from another project (the node part, which is not a dependence of the parser part)
- then only it can go on parsing with the possible addons enabled

[...]

PR #703, opened 2022-09-04
feat: support custom preparser

twitwi:
Targeting https://github.com/slidevjs/slidev/issues/700

- [x] ~~rewrite parser to a single while loop and an explicit state (including the current mode, i.e. what it is currently parsing)~~
- [x] make it extensible by addons
- [x] add addon pseudo-callbacks in useful places ~~(e.g. onSlice)~~ (toplevel all lines (as splitting is already done, and per slide)
- [x] allow easy "wrap" use case (to prepend and append lines to the content)
- [x] integrate tests in the testing infrastructure
- [x] make an initial documentation (a few typical use cases) https://github.com/slidevjs/docs/pull/91/files
- [ ] update the doc with the latest API

I think the core of the PR is fully ok (all parser tests still passing, more tests added for the extensibility).

The question of **clean integration in the project** is to be checked as I'm not fully familiar with its architecture.

### DIFF
--- a/packages/types/src/types.ts
+++ b/packages/types/src/types.ts
@@ -55,4 +55,15 @@ export interface SlidevMarkdown {
   themeMeta?: SlidevThemeMeta
 }
 
+export interface SlidevPreparserExtension {
+  name: string
+  transformRawLines?(lines: string[]): Promise<void>
+  transformSlide?(content: string, frontmatter: any): Promise<string | undefined>
+}
+
+export type PreparserExtensionLoader = (addons: string[], filepath?: string) => Promise<SlidevPreparserExtension[]>
+
+// internal type?
+export type PreparserExtensionFromHeadmatter = (headmatter: any, exts: SlidevPreparserExtension[], filepath: string | undefined) => Promise<SlidevPreparserExtension[]>
+
 export type RenderContext = 'slide' | 'overview' | 'presenter' | 'previewNext'
--- a/packages/parser/src/core.ts
+++ b/packages/parser/src/core.ts
@@ -1,6 +1,6 @@
 import YAML from 'js-yaml'
 import { isObject, isTruthy, objectMap } from '@antfu/utils'
-import type { SlideInfo, SlideInfoBase, SlidevFeatureFlags, SlidevMarkdown, SlidevThemeMeta } from '@slidev/types'
+import type { PreparserExtensionFromHeadmatter, SlideInfo, SlideInfoBase, SlidevFeatureFlags, SlidevMarkdown, SlidevPreparserExtension, SlidevThemeMeta } from '@slidev/types'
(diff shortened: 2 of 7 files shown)


twitwi:
Any feedback is welcome. I'm not perfectly happy with the solution:

- I think the **parser** (and preparser) is though as **an independent module,** so it has few dependencies (if I am wrong on this, then there are probably ways to simplify the problems below)
- Extending the parser requires to load the customizating code from different sources (like addons or the `setup/` folder), which causes two **roadblocks**
    1. the list of **paths to search** for extensions, is **obtained via** `getUserRoots()`, `getAddonRoots()` which are **in the node module**
    2. the list of actually-**enabled addons** is obtained from the **headmatter** (frontmatter of the first slide) so **during the parsing**
 
- I ended up with a **solution** that makes the node module (which has the get...Roots() information) **inject an "addon loader"** inside the parser which **invokes it during parsing**, when the headmatter is closed
- also, I'm not sure the injecting code is in the **best place?**
- I've made the **parsing async,** which is probably **not an issue** as it is only called in an async context already

twitwi:
(not sure what to do about the cypress error which seems unrelated (gpu, dri error))

packages/types/src/types.ts
```diff
@@ -55,4 +55,37 @@ export interface SlidevMarkdown {
   themeMeta?: SlidevThemeMeta
 }
 
+/**
+ * Preparser modes: listing the built-in ones but it is open to extensions.
+ * Modes starting with `:` are pseudo-modes (usually keeping the cursor in place) so that extensions can handle them.
+*/
+export type SlidevPreparserMode = (
```
antfu, line 62: I am not so sure about the complexity here. I would prefer to have simple string transformation for the entire markdown file and then probably for each slide. I am expecting:

```ts
export interface SlidevPreparserExtension {
  name: string,
  transformRawMarkdown?(content: string): string | undefined,
  transformSlide?(no: number, content: string, frontmatter: object): string | undefined
}
```

twitwi: I understand. I need to think more about it: in particular, I'm wondering if it would require that any small transformer re-implement the parser and the serializer.

I think the "injection" ugliness would still be required or do you see another solution?
> - I ended up with a solution that makes the node module (which has the get...Roots() information) inject an "addon loader" inside the parser which invokes it during parsing, when the headmatter is closed
> - also, I'm not sure the injecting code is in the best place?

twitwi:
For the record the old nicely complicated version is archived at https://github.com/twitwi/slidev/tree/preparser-extension-nice-complicated

twitwi:
@antfu if we can find a way to move forward on this, that would be great for (me) writing plugins that could be shared with the community.

PR #709, opened 2022-09-15
feat: add layout of two columns with a header (two-cols-header)

a-h:
I was missing this layout, since it's one I use a lot:

image

In slidev, it looks like this:

image


maurer2:
Hello, looks good. I believe you can omit the `grid-rows-2` or change it two `grid-rows-1`, since the two columns only span one row.

Cheers

a-h:
Thanks, I've rebased off the main branch and updated the commit to keep the commit history tidy.

PR #716, opened 2022-10-05
fix(animation): order map key undefined

rainsia:
I encontered a situation that clicks order map with undefined key. I fixed this bug.

### DIFF
--- a/packages/client/modules/directives.ts
+++ b/packages/client/modules/directives.ts
@@ -45,7 +45,7 @@ export default function createDirectives() {
             elements.value.push(el)
 
           // Set default dir.value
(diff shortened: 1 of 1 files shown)


packages/client/modules/directives.ts
```diff
@@ -45,7 +45,7 @@ export default function createDirectives() {
             elements.value.push(el)
 
           // Set default dir.value
-          if (dir.value === null)
+          if (dir.value === undefined)
```
antfu, line 48:
```suggestion
          if (dir.value == null)
```
Maybe this to capture both null and undefined?

PR #724, opened 2022-10-10
fix: create app template error

alexzhang1030:
Here is the thing:
- package `create-app` invoke build
- build command copy files

but ../../demo/starter/slides.md includes ../../demo/starter/pages/multiple-entries.md:

image

So we must copy `multiple-entries.md` when build.

I create a script to do this work.

Hope will fix this issue :)

### DIFF
--- a/packages/create-app/build.mjs
+++ b/packages/create-app/build.mjs
@@ -1,0 +1,31 @@
+import { copyFileSync, existsSync, mkdirSync, rmSync } from 'node:fs'
+import * as url from 'node:url'
+import { resolve } from 'node:path'
+
+const __dirname = url.fileURLToPath(new URL('.', import.meta.url))
+const __templateDir = resolve(__dirname, 'template')
+const __pagesDir = resolve(__templateDir, 'pages')
+
+const shouldCreatePagesDict = () => existsSync(__pagesDir)
+
+const needCopyFiles = {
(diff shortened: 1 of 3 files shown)


antfu (approved):

dtinth:
Seems like the fix is still unreleased, so creating a slidev today still results in an error.

packages/create-app/build.mjs
```diff
@@ -0,0 +1,33 @@
...
+const __templateDir = resolve(__dirname, 'template')
+const __pagesDir = resolve(__templateDir, 'pages')
+
+const shouldCreatePagesDict = () => !existsSync(__pagesDir)
+
+// key: copy to (relative ./)
+// value: origin (relative ./template)
+const needCopyFiles = {
+  'slides.md': '../../../demo/starter/slides.md',
+  'pages/multiple-entries.md': '../../../demo/starter/pages/multiple-entries.md',
```
luca-moveshelf, line 15: These files don't exist while scaffolding the app with the `create-app` package; the `create-app` folder is effectively the top level in those cases. This is also a security issue as you are effectively trying to access unknown files in a random user's folder

Luca-Terrazzan, line 15: These files don't exist while scaffolding the app with the create-app package; the create-app folder is effectively the top level in those cases. This is also a security issue as you are effectively trying to access unknown files in a random user's folder

PR #728, opened 2022-10-12
fix: adjust height to follow aspect-ratio for printing

### LINKED ISSUE #726: Wrong bottom white space on print
**Describe the bug**
On export to pdf, there is an unwanted white space at the bottom.

**To Reproduce**
1. Set the `aspectRatio` to 16/9
2. Set the `canvasWidth` to 960 (the bug is not present with 1920 or 1280)
3. Create a slide with a colored background (eg: full red)
4. Run the export to pdf command
5. See an unwanted white space at the bottom of the page

 - OS: KDE Neon (Linux)
 - Browser: Chrome 106.0.5249.103
 - Slidev version: 0.36.5
 
 ---
 
 **BONUS**
 I already found the bug on https://github.com/slidevjs/slidev/blob/main/packages/client/internals/Print.vue#L11, we should not add `+ 1` to the calculated height.
 
 I will look for contribute asap.

lomby92:
Resolves #726 

## Error on `main`:
(open the image to see the unwanted space)

main

## Fix
The undesired white space is not present

fixed


packages/client/env.ts
```diff
@@ -7,7 +7,16 @@ import _configs from '/@slidev/configs'
...
+
+// The height must be an integer (apply round) and it should honors the aspect ratio more as possible.
+// Also, to prevent errors on PDF export, we need to increase the height by 1px at time until the real
+// aspect ratio is equal or less than the requested one.
+// Doing this, we will prevent to create empty white pages after all single pages.
+let calculatedHeight = Math.round(slideWidth / slideAspect)
+while (slideAspect < (slideWidth / calculatedHeight))
+  calculatedHeight++
+
+export const slideHeight = calculatedHeight
```
antfu, line 19: Do you mean `Math.ceil`?

lomby92: `calculatedHeight` is already an integer since at line 15 is calculated from Math.round. I'm wrong?

antfu: I mean we can change `Math.round` to `Math.ceil`

lomby92: Changes done, let me know 😉 

Also: since I'm already in, can I ask to enable this PR for the Hacktoberfest?

packages/client/utils.ts
```diff
@@ -6,3 +6,18 @@ export function getSlideClass(route?: RouteRecordRaw) {
...
+ * The height must be an integer (apply round) and it should honors the aspect ratio more as possible.
+ * Also, to prevent errors on PDF export, we need to increase the height by 1px at time until the real
+ * aspect ratio is equal or less than the requested one.
+ * Doing this, we will prevent to create empty white pages after all single pages.
+ */
+export function getSlideHeightRespectingAspectAsPossible(requestedWidth: number, aspectRatio: number) {
+  let calculatedHeight = Math.ceil(requestedWidth / aspectRatio)
+
+  while (aspectRatio < (requestedWidth / calculatedHeight))
+    calculatedHeight++
```
antfu, line 20: I don't get it. With Math.ceil, why this is needed?

lomby92: Leftover, not needed anymore

packages/client/env.ts
```diff
@@ -3,11 +3,13 @@ import { computed } from 'vue'
 import { objectMap } from '@antfu/utils'
 // @ts-expect-error missing types
 import _configs from '/@slidev/configs'
+import { getSlideHeightRespectingAspectAsPossible } from './utils'
 
 export const configs = _configs as SlidevConfig
 export const slideAspect = configs.aspectRatio ?? (16 / 9)
 export const slideWidth = configs.canvasWidth ?? 980
-export const slideHeight = Math.round(slideWidth / slideAspect)
```
antfu, line 10: I think all we need is to change this line. We don't need an extra util on on this.

antfu (approved):

packages/client/env.ts
lomby92: Sorry for the message @antfu: what about to make this PR enabled for Hacktoberfest?

PR #739, opened 2022-10-22
feat: allow to export presenter notes

### LINKED ISSUE #718: export presenter notes only
I'm preparing a talk and would like to print my notes as the room setup does not allow to have a separate window in presenter mode.

Is it possible to add an export of the print notes only, with for instance some information of the slide the notes applies to (eg slide nr, maybe title?)

Thanks!
Pieter-Jan

lomby92:
Resolve #718


antfu:
Looks great, thank you! Would you might to also send a PR to the docs repo mentioning about this feature? Thanks

PR #742, opened 2022-10-24
feat: improve print margin for notes export

lomby92:
Details in the PR code comments

### DIFF
--- a/packages/client/internals/PresenterPrint.vue
+++ b/packages/client/internals/PresenterPrint.vue
@@ -25,7 +25,7 @@ const slidesWithNote = computed(() => rawRoutes
         </div>
       </div>
 
-      <div v-for="(slide, index) of slidesWithNote" :key="index" class="flex flex-col gap-4">
+      <div v-for="(slide, index) of slidesWithNote" :key="index" class="flex flex-col gap-4 break-inside-avoid-page">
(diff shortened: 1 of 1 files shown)


packages/client/internals/PresenterPrint.vue
```diff
@@ -25,7 +25,7 @@ const slidesWithNote = computed(() => rawRoutes
         </div>
       </div>
 
-      <div v-for="(slide, index) of slidesWithNote" :key="index" class="flex flex-col gap-4">
+      <div v-for="(slide, index) of slidesWithNote" :key="index" class="flex flex-col gap-4 break-inside-avoid-page">
```
lomby92, line 28: Avoid to break a single slide block. It's more usable if the slide comments are all together in a single printed page

packages/client/internals/PresenterPrint.vue
```diff
@@ -47,5 +47,19 @@ const slidesWithNote = computed(() => rawRoutes
 <style lang="postcss">
 @page {
   size: A4;
+  margin-top: 1.5cm;
+  margin-bottom: 1cm;
+}
```
lomby92, line 52: Add default print margins, prevent to have text too near the top and the bottom.

PR #750, opened 2022-10-27
fix: currentLayout will no longer be empty

canstand:
When no layout is explicitly configured, the currentLayout value is empty. There are actually two possibilities, `cover` and `default`.


antfu:
Do you mean when you want to read it programmatically?

canstand:
Yes, when I need to create a component in my theme, and the component shows different styles according to the layout. 

Theme users may decide to configure page 1 layout as `intro` or other layouts, currently I can't simply use `$slidev.nav.currentLayout !== 'cover'`.

A simple example is the global Footer (global-bottom.vue), which hides the page number on the cover page:
```vue
<template>
  <footer v-if="$slidev.nav.currentLayout !== 'cover'" class="absolute bottom-0 left-0 right-0 text-center p-2">
    Page {{ $slidev.nav.currentPage }}/{{ $slidev.nav.total }}
  </footer>
</template>
```

PR #759, opened 2022-11-10
feat(export): add basic metatada to exported PDF

Envek:
In https://github.com/slidevjs/slidev/pull/669 `pdf-lib` dependency was added and an option to process exported PDF to add some info to it.

Why not go further and enhance exported PDF a little bit more, adding basic metadata to it?

<details>
<summary>Before:</summary>

Screenshot from 2022-11-10 21-36-14
</details>

<details>
<summary>After (look at window title):</summary>

Screenshot from 2022-11-10 21-34-16
</details>

### DIFF
--- a/packages/slidev/node/export.ts
+++ b/packages/slidev/node/export.ts
@@ -6,6 +6,7 @@ import { parseRangeString } from '@slidev/parser/core'
 import type { SlideInfo } from '@slidev/types'
 import { outlinePdfFactory } from '@lillallol/outline-pdf'
 import * as pdfLib from 'pdf-lib'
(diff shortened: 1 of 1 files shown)


packages/slidev/node/export.ts
```diff
@@ -223,9 +224,18 @@ export async function exportSlides({
       preferCSSPageSize: true,
     })
 
+    // Edit generated PDF: add metadata and (optionally) TOC
+    let pdfData = fs.readFileSync(output)
```
antfu, line 228: Can we use the promises version of `fs`? Thanks

Envek: Sure! Done

PR #776, opened 2022-11-29
refactor: tidy up

Simon-He95:


antfu:
Man, you really should avoid sending such PRs anymore. They are not contributing real value. And please respect the original author's choice/preference.

PR #781, opened 2022-12-01
Update image-right.vue

AlexisRenchon:
!!!!!! Does not work !!!!!!!
It does scale 
But layout is wrong (url left instead of right)


AlexisRenchon:
replaced with #783

Issue #810, opened 2023-01-09
Block code are broken


tonai:
**Describe the bug**
Block code are broken.
See picture:
![blockcode](https://user-images.githubusercontent.com/5246045/211346846-44f84e47-6712-4aab-b58f-1f70afb61863.png)
* Dark mode and light mode are both displayed
* font size is much bigger (compared to previous versions)

**To Reproduce**
Reproducible with the demo slides embedded in the repo:
1. `npm run build`
2. `npm run demo:dev`

**Desktop (please complete the following information):**
 - OS: Ubuntu 22.04
 - Browser: chrome 108.0.5359.94
 - Slidev version: 38.2

tonai:
Introduced by the recent upgrade of `shiki` from `0.11.1` to `0.12.1`

twitwi:
see https://github.com/slidevjs/slidev/issues/799

Issue #799, opened 2022-12-22
code highlighting seems broken on slidev default example


pierreroth64:
**Describe the bug**

When running the slidev default example, code highlighting seems broken.

**To Reproduce**

1. Run `npm init slidev`
2. Go to `http://localhost:3030/4`
3. Use Right Arrow
4. Result: broken layout (code duplicated) and no code highlighting

<img width="769" alt="Capture d’écran 2022-12-22 à 11 19 20" src="https://user-images.githubusercontent.com/4201248/209112789-e4c76ad5-2f32-476c-9317-b763fbd8747a.png">

**Desktop**

 - OS: macOS - Monterey `12.6.2`
 - Node: `v18.12.1`
 - Browser: Firefox `108.0.1` (same issue with Chrome `108.0.5359.124`)
 - Slidev version: `v0.38.2`

⚠️ **IMPORTANT**: works with version `v0.37.1` !

twitwi:
`v0.38.1` works too.
So my best guess for the cause would be the shiki update
https://github.com/slidevjs/slidev/commit/3b79d4765310751e6e04308f79223fb74d0ea4eb#diff-4fc037dd5180dae1af9ac637aa57f5e164f0dd6119ba0da6681646cee8a1afa6R87

twitwi:
Actually shiki has no version `v0.12.1` (the latest being `v0.12.0`)

twitwi:
shiki has no 0.12.1 on github https://github.com/shikijs/shiki/tags
but has a 0.12.1 on npm https://www.npmjs.com/package/shiki

In anycase, both 0.12.0 and 0.12.1 are causing the problem (reverting to 0.11.1 on the latest slidev works).

lili21:
same issue here

PR #811, opened 2023-01-09
fix(shiki): replace pattern to match new version of shiki + fix scrollbars styles

tonai:
Fixes https://github.com/slidevjs/slidev/issues/810
Fixes https://github.com/slidevjs/slidev/issues/799

### DIFF
--- a/packages/client/styles/code.css
+++ b/packages/client/styles/code.css
@@ -1,13 +1,13 @@
+html.dark:root {
+  color-scheme: dark;
+}
+
 html.dark .shiki-light {
   display: none;
 }
 
 html:not(.dark) .shiki-dark {
   display: none;
-}
-
-::-webkit-scrollbar {
-  width: 0px;
 }
 
 .slidev-code-wrapper {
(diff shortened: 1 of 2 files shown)


packages/client/styles/code.css
```diff
@@ -6,10 +10,6 @@ html:not(.dark) .shiki-dark {
   display: none;
 }
 
-::-webkit-scrollbar {
-  width: 0px;
-}
```
antfu, line 11: What's this change is about?

tonai: This code hides the scrollbar.
And I think this is pretty bad, especially when the code overflows.
I you keep this code:
with

If you remove it:
without

packages/slidev/node/plugins/markdown-it-shiki.ts
```diff
@@ -57,10 +57,10 @@ const MarkdownItShiki: MarkdownIt.PluginWithOptions<ShikiOptions> = (markdownit,
       const trimmed = trimEndNewLine(code)
       const dark = _highlighter
         .codeToHtml(trimmed, { lang: lang || 'text', theme: darkModeThemes.dark })
-        .replace('<pre class="shiki"', '<pre class="slidev-code shiki shiki-dark"')
+        .replace('<pre class="shiki', '<pre class="slidev-code shiki shiki-dark')
```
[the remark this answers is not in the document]
antfu: `<` should be escaped in the user code, so it should be fine.

Issue #806, opened 2023-01-05
How can I customize the navigation bar


Krue1:
**Is your feature request related to a problem? Please describe.**
I'm always frustrated when I want to share the slide to other, meanwhile I don't expect others to see my notes.
But the navigation bar always shows the presenter mode button, which can be used to see the notes.

Ps: I packaged the application as an SPA and deploy it to the server to share the slides with others.

**Describe the solution you'd like**
Front matter configures provide options to customize the navigation bar.

PR #819, opened 2023-01-12
feat: option to disable presenter mode

tonai:
Fixes https://github.com/slidevjs/slidev/issues/806

Add front matter configuration to enable or disable the presenter mode (default is enabled)


tonai:
This feature will need some doc update

Issue #692, opened 2022-08-23
Persisted drawings aren't exported to PDF nor hosted site


adrianbienias:
**Describe the bug**
Despite the following setting
```
drawings: 
  persist: true
```
Drawings aren't exported to PDF nor static site.

From the docs, I understand that it should be possible

> The following frontmatter configuration allows you to persist your drawings as SVGs under .slidev/drawings directory and have them inside your exported pdf or hosted site.

https://sli.dev/guide/drawing.html#persist-drawings

**To Reproduce**
1. Create a project `$ npm init slidev`
2. Change drawings config in slides.md to
```
drawings:
  persist: true
```
3. Run the `$ slidev`
4. Draw something on slides
5. Confirm that drawings were saved in `.slidev/drawings`
6. Install `playwright-chromium` https://sli.dev/guide/exporting.html#exporting
7. Export slides to PDF `$ slidev export`
8. See that there are no drawings on exported slides

**Desktop (please complete the following information):**
 - Slidev version: `0.35.5`

lpil:
I am also having this problem when running `npm run build` to compile the presentation to host on my website. When I do this the drawings are absent. There are no errors or warnings in the browser console.

lpil:
Hi robot! Please keep this open.

lpil:
Hi robot, please keep this open.

PR #825, opened 2023-01-16
fix: export drawings in export (if persist is true)

tonai:
Fixes https://github.com/slidevjs/slidev/issues/692

Added a new global variable `__USE_SERVER__` and a new `"export"` mode:

* `mode="dev"` means `__DEV__=true` and `__USE_SERVER__=true`
* `mode="export"` means `__DEV__=false` and `__USE_SERVER__=true`
* `mode="build"` means `__DEV__=false` and `__USE_SERVER__=false`

I didn't forced the export to be in dev mode because in that mod we renders additionnal things like the Monaco editor...etc. we won't use in the export.

### DIFF
--- a/packages/client/state/syncState.ts
+++ b/packages/client/state/syncState.ts
@@ -7,7 +7,7 @@ export function createSyncState<State extends object>(serverState: State, defaul
   let patchingTimeout: NodeJS.Timeout
   let updatingTimeout: NodeJS.Timeout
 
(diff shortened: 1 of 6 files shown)


antfu (approved):

antfu:
We could have it once conflicts are solved

tonai:
Fixed :)

Issue #794, opened 2022-12-16
Chapter Selection (Named Goto)


rawkode:
**Is your feature request related to a problem? Please describe.**

The `g` key, `goto` feature is fantastic - but only if you know the slide number you need to visit.

I'd like to see this extended to offer chapter selection.

**Describe the solution you'd like**

In the frontmatter, we can add `chapter: XYZ` or `name: XYZ` and this gives us a name `XYZ` that we can use to jump around.

`g` input box should then support auto-completing `X` into `XYZ` if there's no ambiguity and go straight to that slide.

**Describe alternatives you've considered**

NA

tonai:
@rawkode here is a draft of a solution for this case.
Can you have a look and tell me what you think ?

rawkode:
@tonai Sorry for the delay. That draft looks awesome.

1. `routeAlias` I think is fine while you gather some feedback from adopters
2. I agree that headers would lead to an overly complicated chapter selection, so let's avoid that.
3. I think a new parameter is a good route forward, but letting people adopt with the routeAlias approach and understanding the vocab people use for chapters could mean less hassle later.

PR #833, opened 2023-01-18
[WIP] Fix/goto route alias

tonai:
Fixes https://github.com/slidevjs/slidev/issues/794

I used the frontmatter `routeAlias` because it was already existing for creating route alias by giving a route a name, so it appears to be very close to that notion of chapter and also because that `routeAlias` frontmatter parameter it not very useful by itself.
But maybe we want to split both feature, tell me.

Also I realized that maybe we can automatically create the autocomplete list by using the slides title and thus we can search for a page without having to specifically give it a `routeAlias` parameter.
But maybe this can leads to having too much items in large presentation...
In that case, maybe we can have a new parameter to enable the automatic chaptering or not.
And maybe we can still use that `routeAlias` frontmatter parameter to override the title for a specific slide (but currently we can override a slide title by using the `title` frontmatter parameter).

So do we want to limit the number of parameters and reuse them for feature that are close ?
Or do we want to create dedicated parameter ?

Also I maybe need to limit the number of autocomplete items (there is no limitation yet).

### DIFF
--- a/packages/client/internals/Goto.vue
+++ b/packages/client/internals/Goto.vue
@@ -1,28 +1,32 @@
 <script setup lang="ts">
-import { computed, nextTick, ref, watch } from 'vue'
-import { go, rawRoutes, total } from '../logic/nav'
+import { computed, ref, watch } from 'vue'
+import { aliases, availablePaths, go } from '../logic/nav'
 import { showGotoDialog } from '../state'
 
 const input = ref<HTMLInputElement>()
 const text = ref('')
 
-const valid = computed(() => {
-  if (text.value.startsWith('/')) {
-    return !!rawRoutes.find(r => r.path === text.value.substring(1))
-  }
-  else {
-    const num = +text.value
-    return !isNaN(num) && num > 0 && num <= total.value
-  }
-})
+const valid = computed(() => isValid(false))
+
+function isValid(strict: boolean): boolean {
+  let path = text.value
+  if (text.value.startsWith('/'))
+    path = text.value.substring(1)
+  if (strict)
+    return availablePaths.value.includes(path)
+  else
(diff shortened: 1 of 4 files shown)


tonai:
Previews:
preview1
preview2

tonai:
Add Height limit for autocomplete list

yannhub:
My team and I are very interested in this new feature. What is the status of this PR?
Thx for your contribution @tonai.

tonai:
It's ready fir review (I forgot to update the PR title)
Currently it is using `routeAlias` in the frontmatter to set a dedicated URL for the slide, that can also be used goto feature.

packages/client/internals/Goto.vue
```diff
@@ -1,74 +1,189 @@
...
+        :class="{ 'text-red-400': !valid && text }"
+        @keydown.enter="goTo"
+        @keydown.escape="close"
+        @keydown.down="focusDown"
+        @keydown.up="focusUp"
+        @input="updateText"
+      >
+    </div>
+    <ul
+      v-if="autocomplete.length > 0"
```
antfu, line 136: Having auto-complete is a good idea. And since we are inferring the title for each slide, maybe we could display them and even do a fuzzy search based on the title as well?

image

tonai: That's what I was proposing in my original message ;)
I agree that it would be nice.
I'm just wondering if this is still relevant for long presentation with a lot of slides.
And in that case maybe we need to add something to distinguish title levels as well in the autocomplete.
Will check.

antfu:
Refactored a bit, now with fuzzy search :)

image

antfu:
Nice work, thanks!

Issue #834, opened 2023-01-18
Nested lists on right-to-left languages are broken


aligit:
**Nesting in lists does not work when dir is right-to-left**
As I speak in Arabic I need to change the direction of the text. The exact same markdown snippet is correctly rendered on [github](https://github.com/aligit/testrtlinmd).

**To Reproduce**
1. Add the following to presentation 
```markdown
<div dir="rtl">
```
3. Create an unordered list of elements as follow:
```markdown
* مقدمه
* معرفی
  * معرفی ۱
  * معفرفی ۲
* نتیجه‌گيری
```
4. You will see that elements 3 and 4 are not nested and they all appear in the same level as other elements.
![56e26bf8-8865-4527-8908-7c1ddee894ac](https://user-images.githubusercontent.com/342484/213261251-1c46fdfe-1bf7-419c-b288-561b6cd744f0.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser: Chrome, Firefox
 - Slidev version: v0.38.3

tonai:
Fixed here https://github.com/slidevjs/slidev/pull/835

For info you can also put the `dir` attribute in the frontmatter (instead of creating a new div):

```md
---
dir: rtl
---
```

PR #835, opened 2023-01-19
fix: styles for rtl languages

tonai:
Fixes https://github.com/slidevjs/slidev/issues/834

### DIFF
--- a/packages/client/styles/layouts-base.css
+++ b/packages/client/styles/layouts-base.css
@@ -6,7 +6,7 @@
   }
 
   h1 {
-    @apply text-4xl mb-4 -ml-[0.05em];
+    @apply text-4xl mb-4;
   }
 
   h2 {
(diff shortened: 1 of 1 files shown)


tonai:
Maybe we need a way to add `dir="rtl"` globally when writting a presentation in a language like arabic ?

antfu:
> Maybe we need a way to add `dir="rtl"` globally when writting a presentation in a language like arabic ?

Maybe `rtl: true` option in the frontmatter?

tonai:
> > Maybe we need a way to add `dir="rtl"` globally when writting a presentation in a language like arabic ?
> 
> Maybe `rtl: true` option in the frontmatter?

Can be, but when we think about it we should probably do the same thing for the `lang` attrbute.
So either:

* we add 2 attributes in the frontmatter, one for `dir` and another one for `lang` (and another one for another futur attribute ?)
* we add a `html` frontmatter attribute where you can add attributes that will be applied to the `html` element. Example:
  ```md
  ---
  html:
    dir: rtl
    lang: ar
  ---
  ```
* we provide a way to customize the `index.html`, in fact this is already possible to add things in the `head` and in the `body`, but we could maybe also use that file to customize `html` attributes.

antfu:
```yaml
htmlAttrs:
  dir: rtl
  lang: ar
```

Probably more specific? In general I think it's a good idea

PR #836, opened 2023-01-19
feat: slide transitions

tonai:
Proposal to add slide transition.

5 built-in transitions for now:
* slide-left
* slide-right
* slide-up
* slide-down
* fade

You can set the transition for all your slides by using the frontmatter `transition` property in the first slide:
```md
---
transition: slide-left
---
```

You can override the transition for a specific slide by again using the frontmatter `transition` property.

By setting the transition for a specific slide, this will determine the transition between this slide and the next one. (using a transition for the last slide will determine the transition between that slide and the "end" slide)

For built-in transitions the system will automatically choose the appropriate transition when going back.
If you set `slide-left` for the third slide it will:
* use `slide-left` when going from 3 to 4
* use `slide-right` when going from 4 to 3

You can write your own transition because underneath it uses `TransitionGroup`.
Thus the custom name you set in the `transition` property will be the name used for `TransitionGroup`.

In that case you might want to define the backward transition.
You can do it by using this format:
```md
---
transition:
  forward: slide-and-fade-left
  backward: slide-and-fade-right
---
```


tonai:
I set it as a draft to discuss it but the current implementation is already working

antfu:
Awesome work! Sorry I missed your PR and happened to work on the same thing together with @posva. Your approach is interesting and I think it's better than my current one in many aspects. I will try to see how we can merge the two ideas together.

antfu:
It's working perfectly! Thanks a lot for this!

tonai:
> It's working perfectly! Thanks a lot for this!

glad I could help :)

PR #837, opened 2023-01-20
fix: toc styles

tonai:


antfu:
I am worried about if this would make theming and custom style harder. Ideally, I think we should avoid using `!important`

tonai:
I chose to use `!important` because I didn't see any reason we would want to have a paragraph in the toc in fact.
But it is also ok to only increase the selector specificity.

tonai:
I also fixed the lint error from the CI that comes from https://github.com/slidevjs/slidev/pull/840
It seems the CI does not run when the PR comes from a fork.

Issue #317, opened 2021-07-25
[Feature Request] Allow build with multiple markdown entries


Shigma:
**Is your feature request related to a problem? Please describe.**
My team uses slidev to present slides every week, and our project directory looks like this:
```
root
  components
  slides
    20210710.md
    20210717.md
    20210724.md
  package.json
```
I want to build all this slides into a SPA but it seems that slidev only supports building with single entry.

**Describe the solution you'd like**
```
slidev build [file1] [file2]
slidev build [directory]
```

**Describe alternatives you've considered**
N/A

antfu:
Why not just running the script multiple times?

Shigma:
> Why not just running the script multiple times?

Because the assets that slidev generates could have been shared, but it didn't.

```
dist
  assets
    vendor.xxxxxx.js
  index.html
```

What I want is

```
dist
  assets
    vendor.xxxxxx.js (actual "vendor")
    1.xxxxxx.js (for 20210710.md)
    2.xxxxxx.js (for 20210717.md)
    3.xxxxxx.js (for 20210724.md)
  20210710.html
  20210717.html
  20210724.html
```

And each of the generated HTML file have something in common.

antfu:
I guess this could be done by providing multiple input entries to Vite's build. PR Welcome.

twitwi:
cross referencing https://github.com/slidevjs/slidev/issues/479 https://github.com/slidevjs/slidev/issues/505

LeonLiuY:
> I made an implementation that is working You can access at: https://github.com/gabriel-del/slidev All feedbacks are welcome

This no longer works with `@slidev/cli v0.32.3` as it fails to resolve package.json under the `presentations` folder.

gabriel-del:
It works to my friends, it takes some time to buid, maybe 3min, and maybe you need to use npm instead of yarn.
Did you get any error mesage?

LeonLiuY:
Yes I'm using npm.

```
> ./slidev.sh build

Error: Failed to resolve package "/*******/my-slides/presentations/package.json"
    at resolveImportPath (/*******/my-slides/node_modules/@slidev/cli/dist/chunk-JWGORJY6.js:121:11)
    at getPackageJson (/*******/my-slides/node_modules/@slidev/cli/dist/chunk-N7XL66RP.js:9515:72)
    at getAddons (/*******/my-slides/node_modules/@slidev/cli/dist/chunk-N7XL66RP.js:9518:33)
[...]

PR #844, opened 2023-01-23
feat: update cli to accept multiple entry files

tonai:
Commands that now accept multiple entry files:
* build
* format
* export
* export-notes

Works like this:
```bash
slidev export slides/1.md slides/2.md
```
Or by using terminal wildcards for example:
```bash
slidev export slides/*.md
```

We need to discuss the following points:

1. Some option arguments are only compatible with one entry file.
    Like `output` for `export` end `export-notes`.
    And like `base` for `build`.
    Maybe we should not take into account these options when `entry.length > 1`.
2. Currently when building with multiple entry files, it will create a subfolder in the `dist` directory for each entry file.
    For example `slidev export slides1.md slides2.md` will generate two build folders `dist/slides1` and `dist/slides2`.
    But because some files are the same between the builds maybe we can still build in the same `dist` directory and only rename conflicting files like `index.html` to have `dist/slides1.html` and `dist/slides2.html`.

For now the default "dev" command does not accept multiple entry files.
Maybe it can be possible by updating the routing and all virtual packages but this might require a lot of work (maybe we can do it in another PR).

### DIFF
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -237,7 +237,7 @@ cli.command(
 )
 
 cli.command(
-  'build [entry]',
+  'build [entry..]',
   'Build hostable SPA',
   args => exportOptions(commonOptions(args))
     .option('watch', {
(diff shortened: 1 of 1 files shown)


tonai:
Partially fixes https://github.com/slidevjs/slidev/issues/317

tonai:
This can be reviewed.
It does not introduce any breaking changes.

antfu (approved):

Issue #764, opened 2022-11-19
Custom JS Vue components are not rendered


vibl:
**Describe the bug**
Custom JS, JSX and TSX Vue components are not rendered.

**To Reproduce**

1. In a freshly installed Slidev project (with `npm init`), add a component `component/MyComponent.js`
```
export default {
  render(){
    return "Hello World"
  }
};
```

3. In `slides.md` , add
```
<MyComponent />
```
The component does not render, and three errors are thrown in the Chrome Devtools console, that start with:
```
runtime-core.esm-bundler.js:40 [Vue warn]: Failed to resolve component: MyComponent
If this is a native custom element, make sure to exclude it from component resolution via compilerOptions.isCustomElement. 
```

Same with JSX and TSX components.

**Desktop (please complete the following information):**
 - OS: Kubuntu
 - Browser: Chrome 
 - Slidev version: 0.36.11

vibl:
I found the cause: 
https://github.com/slidevjs/slidev/blob/cc671b400b9401b56c8be8440f9b688bc346f406/packages/slidev/node/plugins/preset.ts#L102

This contradicts the documentation: https://sli.dev/custom/directory-structure.html#components

vibl:
After adding ` 'jsx', 'tsx'` in the array mentioned above, I get an error "ReferenceError: React is not defined" when I use JSX in an .jsx or .tsx file.

vibl:
It works after adding installing `@vitejs/plugin-vue-jsx` with a  `vite.config.js` file as per https://github.com/vitejs/vite/blob/main/packages/plugin-vue-jsx/README.md

vibl:
And to include the missing file extensions, the complete `vite.config.js` is:
```
import vueJsx from '@vitejs/plugin-vue-jsx';
import { defineConfig } from 'vite';

export default defineConfig({
  plugins: [
    vueJsx()
  ],
  slidev: {
    components: {
      extensions: ["vue", "md", "js", "ts", "jsx", "tsx"]
    },
  },
});
```

tonai:
Maybe can you propose a PR to add support of jsx and tsx files ?

PR #846, opened 2023-01-24
feat: support js jsx tsx components

tonai:
Fixes https://github.com/slidevjs/slidev/issues/764


antfu:
I think I resisted it to `.ts` on purpose to advocate the usage of TS. But anyway no harm to support more :)

PR #856, opened 2023-02-01
Improve performance by compressing in parallel 🚀

sanjaiyan-dev:
Improving the performance by running the independent `async` functions in parallel ✈️.

_Extremely sorry if I made any mistakes :(_

### DIFF
--- a/packages/slidev/node/export.ts
+++ b/packages/slidev/node/export.ts
@@ -400,8 +400,7 @@ export async function exportSlides({
     await genPagePng()
   }
   else if (format === 'md') {
-    await genPagePng()
-    await genPageMd(slides)
+    await Promise.all([ genPagePng(), genPageMd(slides)])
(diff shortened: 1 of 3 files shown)


antfu:
I am not against it, but do you see any real-world improvements? I'd judge those are not the bottlenecks in terms of performance.

packages/slidev/node/export.ts
```diff
@@ -400,8 +400,7 @@ export async function exportSlides({
     await genPagePng()
   }
   else if (format === 'md') {
-    await genPagePng()
-    await genPageMd(slides)
+    await Promise.all([ genPagePng(), genPageMd(slides)])
```
tonai, line 403: The `genPageMd` function reads the `output` dir where the images are exported to then generate the md file, so we can't process them in parallel

Issue #798, opened 2022-12-20
Ability to change image's backgroundSize on `image` layout different than `cover`


alpcoker:
While using screenshots in presentation, images with a different proportion (for example a vertical image) than the slide ratio are cropped. It is the side effect of using `backgroundSize: cover` in the [handleBackground function](https://github.com/slidevjs/slidev/blob/main/packages/client/layoutHelper.ts#L31).

**To Reproduce**

1. Go to https://stackblitz.com/edit/slidev-image-custom-size
2. The first slide is a default image layout. You can see the image is cropped compared to the latter slide.
3. The second slide is my custom solution with a custom layout. You can see all of the image.

I think some people will need to show their images without any cropping. Three options come to my mind that I would love to be involved in. 

1. Adding another layout with strict naming `image-contain`
2. Customizable layout, with a prop `image-custom-size` 
3. Adding an optional property to the existing image layout. By preserving default `cover` behavior. 

I have solved this requirement in this environment: https://stackblitz.com/edit/slidev-image-custom-size by following these;

1. Created a clone of the image layout
2. Add `backgroundSize` prop to the layout and in my slide set this as `contain`
3. Send this prop value clone of the [`handleBackground`](https://github.com/slidevjs/slidev/blob/main/packages/client/layoutHelper.ts#L12) function with an existing default parameter `cover`

If any of my proposals are suitable, I would love to open a PR.

asm0dey:
Yes please, I need it!

tonai:
Yes please propose a PR.

asm0dey:
up

alpcoker:
Hey @antfu you left an initial comment on PR: https://github.com/slidevjs/slidev/pull/862, we can discuss continuing here. I [commented](https://github.com/slidevjs/slidev/pull/862#issuecomment-1418249626) on PR but you may miss it. As it has been a long time, I wanted to make this topic reactive. 

Before discussing proposes. The first question will be; 

Do we need that feature? Seems like two other people in this issue also want that feature. If not, we can close.

**Proposes**

[...]

PR #862, opened 2023-02-03
Custom backgroundSize for image layouts (#798)

alpcoker:
fixes https://github.com/slidevjs/slidev/issues/798

With this additional prop to image layouts (image, image-left, image-right), we can now able give custom css [background-size](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size). The default will remain `cover` 

My main motivation was to be able to use `contain`. Because I wasn't able to fit vertical photographs, let's cover all the different needs.  

Example usages:

```
---
layout: image
image: https://iili.io/HzdAhJV.png
backgroundSize: contain
---
```

```
---
layout: image-left
image: https://iili.io/HzdAhJV.png
backgroundSize: 20em 70%
---
```


antfu:
I honestly feel this is a bit too specially handled. When you want to customize it, there are many more other styles to configure as well. I'd suggest we expose classes or styles to the image. Maybe:

```md
---
imageStyle:
  backgroundPosition: center
  backgroundSize: etc
---
```

alpcoker:
I love the simplicity of the layout `image[-left|-right].vue` by just setting the image URL. The only problem is that the existing `background-size: cover` is not suitable for some usages because it scales.

I also agree that this extra property seems very specific, but I think the `background-size` property would be the most used `background-*` property after `background-image`. I am not sure that other properties would be used that much.

- I tried your opinion for an ability to set other background properties with a named `image-custom` layout.
- I also wanted to try another layout that's `background-size: contain` as a static. The drawback is to create for also left & right layout too.

To compare & discuss 3 options temporarily, I have changed the demo project's `slide.md`

alpcoker:
Hello @antfu. As I see conflicts in my PR, I rebased and for the demo I added `pnpm demo:feat` (just image layout demos created in slides-feat.md) to be easily review. I also added the slide screenshots.

demo/starter/slides-feat.md
```diff
@@ -0,0 +1,31 @@
+---
```
antfu, line 1: The starter is meant to be an example showcase and a playground. For tests like this it's better to move to test fixtures.

packages/client/layouts/image-custom.vue
```diff
@@ -0,0 +1,50 @@
+<script setup lang="ts">
```
antfu, line 1: I am not sure about this layout. I am concerned about the slope effect that we will end up with all CSS rules. I'd recommend creating your own local layouts for your specific needs.

antfu (changes requested):

packages/client/layouts/image-custom.vue
alpcoker: Sorry for accidentally deleting the branch, fix that & I have [opened new PR](https://github.com/slidevjs/slidev/pull/862) so we can continue from there.

demo/starter/slides-feat.md
alpcoker: Sorry for accidentally deleting the branch, fix that & I have [opened new PR](https://github.com/slidevjs/slidev/pull/862) so we can continue from there.

Issue #865, opened 2023-02-06
Timeout while rendering PDF


asm0dey:
**Describe the bug**

**To Reproduce**
1. Clone github.com/asm0dey/clean-code-101
2. run `DEBUG=pw:api  pnpm run export-pdf`
3. See something like 

```
❯ DEBUG=pw:api pnpm run export-pdf

> kotlin-spring@ export-pdf /home/finkel/work_self/presentations/clean-code-101
> slidev export --with-clicks --format pdf --output dist/index.pdf



  ●■▲
  Slidev  v0.40.2 

  theme   @slidev/theme-apple-basic
  entry   /home/finkel/work_self/presentations/clean-code-101/slides.md


  pw:api => browserType.launch started +0ms
  pw:api <= browserType.launch succeeded +207ms
  pw:api => browser.newContext started +2ms
  pw:api <= browser.newContext succeeded +700ms
  pw:api => browserContext.newPage started +0ms
  pw:api <= browserContext.newPage succeeded +746ms
  ● rendering...  pw:api => page.goto started +3ms
  pw:api navigating to "http://localhost:12445/print?print=clicks", waiting until "networkidle" +2ms
  pw:api   "commit" event fired +143ms
  pw:api   navigated to "http://localhost:12445/print?print=clicks" +0ms
  ■ rendering...  pw:api   navigated to "http://localhost:12445/print?print=clicks" +417ms
  ▲ rendering...  pw:api   navigated to "http://localhost:12445/print?print=clicks" +124ms
  pw:api   "domcontentloaded" event fired +1ms
  ◆ rendering...  pw:api   "load" event fired +368ms
  ◆ rendering...  pw:api <= page.goto failed +29s
page.goto: Timeout 30000ms exceeded.
=========================== logs ===========================
navigating to "http://localhost:12445/print?print=clicks", waiting until "networkidle"
============================================================
    at go (/home/finkel/work_self/presentations/clean-code-101/node_modules/.pnpm/@slidev+cli@0.40.2_playwright-chromium@1.30.0/node_modules/@slidev/cli/dist/export-BRPXINGP.js:143:16)
    at genPagePdfOnePiece (/home/finkel/work_self/presentations/clean-code-101/node_modules/.pnpm/@slidev+cli@0.40.2_playwright-chromium@1.30.0/node_modules/@slidev/cli/dist/export-BRPXINGP.js:226:11)
[...]

PR #882, opened 2023-02-13
fix: timeout export with windicss

tonai:
Fixes https://github.com/slidevjs/slidev/issues/865

Remove import of `virtual:windi-devtools` causing pending call to `/@windicss-devtools-update` to hang indefinitely 
windi


tonai:
In fact this call also seems to hang indefinitely in dev but I created an other issue for that https://github.com/slidevjs/slidev/issues/883
It was quicker to just disable the Design in DevTools feature in the export as it is completely useless in that case.

Issue #898, opened 2023-03-06
Server ref sync "crash" when using windicss


tonai:
**Describe the bug**
When importing a markdown file (using `src` in frontmatter block) within a markdown that is itself imported, then the presenter/view sync stop to work when reaching the slides belonging to the "deep" markdown file.

**To Reproduce**
1. Create main `slides.md`:
```md
---
theme: seriph
---

# Formation

---
src: ./chapters/part1.md
---
```
2. create `chapters/part1.md`:
```md
## Part 1

blablabla

---
src: ./part1/part11.md
---
```
3. create `chapters/part1/part11.md`:
```md
### Part 1.1

blablabla

blablabla

---

#### Part 1.1.1

blablabla

blablabla

blablabla

---

#### Part 1.1.2

blablabla

blablabla

blablabla

```

**Desktop (please complete the following information):**
 - OS: Ubuntun 22.04
 - Browser: Chrome 109.0.5414.119 
 - Slidev version: 0.40.3

tonai:
When the problem appear, try to refresh the browser is long and an error occur when trying to getch `/@windicss-devtools-update`.

The problem does not appear anymore when switching to unocss.

Issue #897, opened 2023-03-04
Presenter Mode not Syncing with Play Mode Instances


samkasbawala:
**Describe the bug**
The presenter view seems to not be updating the other instances of the same presentation. I have also had some occurrences where the presenter mode version of the slides is not displaying the correct UI. In that I still have the option to record my screen and that there is no option to witch back to play mode. For what ever reason, when I test the demo slidev slides, everything seems to work, but I highly doubt it's user error since the slide content should not be affecting the UI.

**To Reproduce**

I created a really basic `slides.md` document:
```markdown
---
theme: default
---

# This is a test

---

# This is a second slide

Here is some content

---

# Here is a third slide

Here is some more content on the third slide
```

When I start the slidev server and open the slides in presenter view and have play mode next to it. Navigating through the slides in presenter mode doesn't sync with what's on play mode.

![image](https://user-images.githubusercontent.com/65199913/222921025-17bf44ba-77ec-463b-868f-bdc2dc0fabd3.png)

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser: Chrome
 - Slidev version: 0.40.3

samkasbawala:
Here is an example of being in presenter mode in another project. You can see that the option to enter presenter mode is still there, despite already being in presenter mode, with no option to return to play mode.

![image](https://user-images.githubusercontent.com/65199913/222921632-f9c29680-3c11-4f6a-8cab-d9eaab4993f7.png)

MMMikeM:
I'm experiencing the same thing. Remote presenter works a little better but still dies

tonai:
Does it work when you switch to unocss ?
You can switch from windi to unocss by adding in the frontmatter block of the first slide:
```
---
css: unocss
---
```

samkasbawala:
yeah, seems like presenter mode worked when I switched to unocss.

Raniz85:
I've got the same issue and I'm using unocss already.

The cursor shows up in both windows but navigation and drawings doesn't transfer.


https://user-images.githubusercontent.com/4972103/223632604-45a3e5eb-2531-4f30-9d9f-593988431915.mp4

Raniz85:
Here's another video showing the console displaying incoming  cursor movements but nothing when I change slides.


https://user-images.githubusercontent.com/4972103/223633270-e81ef8ec-7d98-4194-8f97-3c9c95512930.mp4

tonai:
[...]

Issue #894, opened 2023-02-24
Background image in style might not be loaded in dev mode


jgosmann:
**Describe the bug**
When importing an image in a Vue component and setting it as background image in CSS via a `style` attribute, it might not be loaded in dev mode.

**To Reproduce**
The bug seems to be somewhat timing dependent and might not always be reliably reproducible. But I did manage to put together a quite minimal repository that allows me to reproduce the bug reliably: jgosmann/slidev-img-load-bug-reproducer

You should be able to observe the bug if you follow these steps:

1. Checkout the repository
2. Run `npm install`
3. Run `npm run dev`
4. Flip through the slides quickly until you reach the last slide before the "END" slide. For me it shows a blank slide instead of loading the "B" image.

Note that the intermediate slides are somehow relevant. If I remove them, I cannot reproduce the bug. If I use less, the bug becomes less reliable to reproduce. In fact, it seems to be related to timing: if I flip quickly through the slides, it still reproduces, but if I do it more slowly, the bug disappear.

Also note, that reloading the "B" slide will load the image. But then flipping back to the "A" slide that image will be missing. 

**Desktop (please complete the following information):**
 - OS: macOS 13.2.1
 - Browsers:
   - Firefox 110.0
   - Chrome 110.0.5481.100
 - Slidev version:
    - Earliest Slidev version reproducing the bug: **0.40.1**
    - Also reproduces with the latest Slidev version which is **0.40.3** at time of writing
    - Does **not** seem to reproduce with Slidev version 0.40.1
    - Only occurs in dev mode. Building a static version and using that is fine.
    - Only occurs with windicss (does not occur with unocss)

**Some further comments**
I looked through the changes between Slide 0.40.0 and 0.40.1 which aren't very many. I can only imagine ee634caabbe36fc9e2fc44419c933f3a0717806d being the cause, though it is not clear to me how. (Neither, it is clear why those plugins where moved to the end to me.) Or maybe it is some weird timing issue that already existed before, but was surfaced through the other change?

It is also somewhat weird that the style is properly applied, when checking in the dev tools. But the browser says it couldn't load the image. However, when clicking the URL it just loads fine. In the network tab, one can see the request (among others) as "pending".

![image](https://user-images.githubusercontent.com/850157/221254471-97932573-8a47-4698-84b3-a6b1730a44a7.png)

jgosmann:
I also made a screen recording:
[...]

PR #899, opened 2023-03-07
fix: Windi CSS plugin must be before ServerRef plugin

tonai:
Fixes https://github.com/slidevjs/slidev/issues/898
Maybe also:
* https://github.com/slidevjs/slidev/issues/897
* https://github.com/slidevjs/slidev/issues/894

### DIFF
--- a/packages/slidev/node/plugins/preset.ts
+++ b/packages/slidev/node/plugins/preset.ts
@@ -148,6 +148,13 @@ export async function ViteSlidevPlugin(
       })
       : null,
 
+    // Must be before ServerRef.
+    config.css === 'none'
+      ? null
+      : config.css === 'unocss'
+        ? await createUnocssPlugin(options, pluginOptions)
+        : await createWindiCSSPlugin(options, pluginOptions),
+
     ServerRef({
       debug: process.env.NODE_ENV === 'development',
       state: {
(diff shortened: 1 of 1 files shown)


antfu:
Did you find the root cause of this? Sounds like it could be a bug of sever-ref?

tonai:
> Did you find the root cause of this? Sounds like it could be a bug of sever-ref?

No, so far I've spent some time trying to figure out what was going on and then realized that moving up the plugin fixes the problem.
But yeah there is probably some kind of compatibility issue between ServerRef and Windi plugins.
I also think that it can be great to find the root issue, but I don't think I will have the time to investigate it now.

tonai:
@antfu maybe it is missing a call to `next` here https://github.com/antfu/vite-plugin-vue-server-ref/blob/main/src/index.ts#L35 ?

tonai:
Yes will do

tonai:
Here is the PR: https://github.com/antfu/vite-plugin-vue-server-ref/pull/3

antfu:
With the fix of https://github.com/antfu/vite-plugin-vue-server-ref/pull/3 this might not be needed right?

tonai:
> With the fix of [antfu/vite-plugin-vue-server-ref#3](https://github.com/antfu/vite-plugin-vue-server-ref/pull/3) this might not be needed right?

Yes just tested and it seems ok with vite-plugin-vue-server-ref 0.3.1

Issue #699, opened 2022-09-01
`<video>` and presenter mode


pomber:
**Is your feature request related to a problem? Please describe.**
When you include a `<video>` in your slide you have to either:
- `autoPlay`: which has a few problems (the video starts autoplaying in the presenter next slide preview, it also starts autoplaying in the play view before the slide is selected)
- `autoPlay="false"`: you can't control the video from the presenter mode 

**Describe the solution you'd like**
Maybe a built-in Video component with a smarter `autoPlay` and controls that sync the video playback between presenter and play mode.

---

Related: #339 #124

antfu:
Sounds like a good idea. Open for PR :)

pomber:
> Sounds like a good idea. Open for PR :)

I may give it a try after I finish my talk. Let me know if there's a similar component or something I could use as a reference.

twitwi:
For the implementation, I would (first) consider something like https://css-tricks.com/a-few-functional-uses-for-intersection-observer-to-know-when-an-element-is-in-view/ (see use case 2 for instance).

PR #905, opened 2023-03-10
feat: add Video built-in component

tonai:
Fixes https://github.com/slidevjs/slidev/issues/699

Usage:
```html
<SlidevVideo autoPlay="resume" muted controls width="250">
  <source src="/flower.webm" type="video/webm">
  Download the <a href="/flower.webm">WEBM</a>
</SlidevVideo>
```

props:
* `autoPlay` (`boolean | 'once' | 'resume' | 'resumeOnce'`): automaticaly play the video when it becomes active (within current slide or within current active click element)
  * `true` or `'once'`: if the video is stopped, it does not restart.
  * `'resume'`: video resumes if stopped and then reactivated. If the video has reached the end, then it will start over from the beginning when it is reactivated.
  * `'resumeOnce'`: video resumes if stopped and then reactivated, but only if it has not already reached the end.
* `autoPause` (`'slide' | 'click'`): automaticaly stop the video
  * `'slide'`: when the video is not in the current slide
  * `'click'`: when the video is not in the current active click element
* `autoReset` (`'slide' | 'click'`): automaticaly reset the video
  * `'slide'`: resets only when the current page containing the video becomes active
  * `'click'`: resets every time the current click element containing the video becomes active


antfu:
I think we would better name it in two words to avoid confusing it with the native `<video` element. Like `<SlidevVideo` maybe?

Issue #572, opened 2022-05-08
Support nice diffs with Monaco


illright:
**Is your feature request related to a problem? Please describe.**
Regular syntax highlighters make you choose between highlighting language syntax or diff lines (with `lang=diff`). Even then, the highlights are not very prominent, certainly not as much as they could be with [Monaco's built-in diff viewer](https://microsoft.github.io/monaco-editor/playground.html#creating-the-diffeditor-navigating-a-diff). 

**Describe the solution you'd like**
I'd like to be able to specify in Markdown that I want to render a diff view in Monaco with _before_ and _after_ source codes.

The syntax could look like this:

    ```ts {monaco} diff-before=my-snippet1
    console.log('Hello, World!');
    ```

    ```ts {monaco} diff-after=my-snippet1
    console.log('Goodbye, World!');
    ```

where `my-snippet1` is a unique identifier for a pair of code blocks describing a diff.

**Describe alternatives you've considered**

* Rendering diffs using regular syntax highlighters — meh
* Using the customization features of Monaco in Sli.dev — not enough
* Attach source code diffs as vector images — meh

illright:
I still want this feature

PR #909, opened 2023-03-13
feat: diff support for monaco editor

tonai:
Fixes https://github.com/slidevjs/slidev/issues/572

monaco-diff-0

With line numbers:
monaco-diff

### DIFF
--- a/packages/slidev/node/plugins/markdown.ts
+++ b/packages/slidev/node/plugins/markdown.ts
@@ -97,6 +97,13 @@ export async function createMarkdownPlugin(
 
 export function transformMarkdownMonaco(md: string) {
   // transform monaco
+  md = md.replace(/^```(\w+?)\s*{monaco}\s*?({.*?})?\s*?\n([\s\S]+?)^~~~\s*?\n([\s\S]+?)^```/mg, (full, lang = 'ts', options = '{}', code: string, diff: string) => {
+    lang = lang.trim()
+    options = options.trim() || '{}'
+    const encoded = base64.encode(code, true)
+    const encodedDiff = base64.encode(diff, true)
+    return `<Monaco :code="'${encoded}'" :diff="'${encodedDiff}'" lang="${lang}" v-bind="${options}" />`
+  })
   md = md.replace(/^```(\w+?)\s*{monaco}\s*?({.*?})?\s*?\n([\s\S]+?)^```/mg, (full, lang = 'ts', options = '{}', code: string) => {
     lang = lang.trim()
     options = options.trim() || '{}'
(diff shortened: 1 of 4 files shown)


packages/slidev/node/plugins/markdown.ts
```diff
@@ -97,6 +97,13 @@ export async function createMarkdownPlugin(
 
 export function transformMarkdownMonaco(md: string) {
   // transform monaco
+  md = md.replace(/^```(\w+?)\s*{monaco}\s*?({.*?})?\s*?\n([\s\S]+?)^~~~\s*?\n([\s\S]+?)^```/mg, (full, lang = 'ts', options = '{}', code: string, diff: string) => {
```
antfu, line 100:
```suggestion
  md = md.replace(/^```(\w+?)\s*{monaco-diff}\s*?({.*?})?\s*?\n([\s\S]+?)^~~~\s*?\n([\s\S]+?)^```/mg, (full, lang = 'ts', options = '{}', code: string, diff: string) => {
```

I am thinking of integrating retypewriter and other tools later with similar syntax (multiple-step blocks). I think we might be better to be explicit here.

antfu (commented):
Great! Can we have some tests for it?

packages/slidev/node/plugins/markdown.ts
tonai: What do you think about the separator `~~~` inside the code block ?

antfu: I was thinking to introduce a syntax for that indeed and I think I end up with the same as your idea of using `~~~`

lfire:
> Fixes #572
> 
> monaco-diff-0
> 
> With line numbers: monaco-diff
> 
> Use `~~~` in the same monaco code block to switch to diff mode:
> 
> ```
> <div>
>   foo
> </div>
> ~~~
> <div>
>   bar
> </div>
> ```

@tonai how to set the options of monaco side by side. In my case, the options like: 
```
```ts {monaco-diff} { editorOptions: { lineNumbers: 'on' }, diffEditorOptions: { renderSideBySide: true, useInlineViewWhenSpaceIsLimited: false } }
export const qsConfig: qs.IParseOptions = {
  allowDots: true,
  depth: 5,
  parameterLimit: 1000,
  arrayLimit: 1000,
  parseArrays: true,
  plainObjects: true,
  strictNullHandling: true,
  ignoreQueryPrefix: true,
};
~~~
export const qsConfig: qs.IParseOptions = {
  parameterLimit: 1000,
  arrayLimit: 1000,
  allowDots: true,
  parseArrays: true,
  depth: 5,
  plainObjects: true,
  strictNullHandling: true
};
```
```
the monaco not working in side by side.  
<img width="1207" alt="image" src="https://github.com/slidevjs/slidev/assets/2528719/4e243796-8622-4c5f-991c-c914e563628c">

thx for your help.

**Comment by lfire** [comment]
sorry, my config is error. thx!
The correct configuration is：
```
ts {monaco-diff} { editorOptions: { lineNumbers: 'on', renderSideBySide: true, useInlineViewWhenSpaceIsLimited: false } }
```

Issue #932, opened 2023-03-30
Slidev 0.40.5 breaks static assets with customized public directory


Hydragyrum:
**Describe the bug**
I have a presentation created with Slidev 0.40.3 with a base path set, and a static assets directory set to `/static` via the vite config. When building it works fine. 

Updating to 0.40.5, with zero changes to code, nor config, the build fails as it can no longer find my static assets at build time.
Logs as follows:

```
$ pnpm build --base /dungeon-master-prez/ --out public
> dungeon-master@ build /builds/Hydragyrum/dungeon-master-prez
> slidev build "--base" "/dungeon-master-prez/" "--out" "public"
  ●■▲
  Slidev  v0.40.5 
  theme   slidev-theme-hydra-conf
  entry   /builds/Hydragyrum/dungeon-master-prez/slides.md
vite v4.2.1 building for production...
transforming...
/fonts/BigShouldersInlineDisplay.ttf referenced in /builds/Hydragyrum/dungeon-master-prez/styles/base.css didn't resolve at build time, it will remain unchanged to be resolved at runtime
/fonts/Nunito.ttf referenced in /builds/Hydragyrum/dungeon-master-prez/styles/base.css didn't resolve at build time, it will remain unchanged to be resolved at runtime
/fonts/FiraCode.ttf referenced in /builds/Hydragyrum/dungeon-master-prez/styles/base.css didn't resolve at build time, it will remain unchanged to be resolved at runtime
✓ 87 modules transformed.
✓ built in 3.00s
[vite]: Rollup failed to resolve import "/img/devoxx-france-light.png" from "/@slidev/slides/1.md".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
Error: [vite]: Rollup failed to resolve import "/img/devoxx-france-light.png" from "/@slidev/slides/1.md".
This is most likely unintended because it can break your application at runtime.
If you do want to externalize this module explicitly add it to
`build.rollupOptions.external`
    at onRollupWarning (file:///builds/Hydragyrum/dungeon-master-prez/node_modules/.pnpm/vite@4.2.1/node_modules/vite/dist/node/chunks/dep-79892de8.js:46368:19)
    at onwarn (file:///builds/Hydragyrum/dungeon-master-prez/node_modules/.pnpm/vite@4.2.1/node_modules/vite/dist/node/chunks/dep-79892de8.js:46138:13)
[...]

PR #935, opened 2023-03-31
fix: allow vite config override

tonai:
Fixes https://github.com/slidevjs/slidev/issues/932


Dtdhamu:
> Fixes https://github.com/slidevjs/slidev/issues/932

PR #944, opened 2023-04-09
feat: support mermaid `mindmap` diagrams

Himself65:


packages/client/modules/mermaid.ts
```diff
@@ -23,7 +23,7 @@ export function renderMermaid(encoded: string, options: any) {
   })
   const code = decode(encoded)
   const id = nanoid()
-  const svg = mermaid.render(id, code) as string
+  const { svg } = await mermaid.render(id, code)
```
Himself65, line 26: in 10.x, there're only async function to render

antfu (approved):
~~感谢男明星！(~~

PR #946, opened 2023-04-09
chore: update `monaco-editor`

vaakian:
I am writing a slide about TypeScript tricks, but the ts version in `monaco-editor@^0.33.0` is `4.5.5`(which is released almost 2 years ago), that lacks a lot of features.
And the latest `monaco-editor@^0.37.x` has ts `5.0.3` built-in.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -32,7 +32,7 @@
     "js-yaml": "^4.1.0",
     "katex": "^0.16.4",
     "mermaid": "^10.1.0",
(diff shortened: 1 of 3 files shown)


antfu:
Have you been able to verify if it works?

vaakian:
Sure, I manually replaced the `0.33.x` version in the `node_modules` with the `0.37.x`, it just works perfectly fine and I was able to use the latest typescript feature in the slide.

PR #948, opened 2023-04-09
fix: prevent ToC links to be blocking element

pascalheidmann:
In Firefox the built-in `TOC` element gets rendered with implicit blocking `div` + `p` which results in follwing problem:
slidev-toc-display-block

Chromium seems to think otherwise of those blocking elements. Nonetheless, resetting those element styles fixes the problem both for Firefox and Chromium browsers.

### DIFF
--- a/packages/client/builtin/TocList.vue
+++ b/packages/client/builtin/TocList.vue
@@ -52,4 +52,7 @@ const classes = computed(() => {
 .slidev-layout .slidev-toc-item p {
   margin: 0;
 }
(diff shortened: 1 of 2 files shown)


packages/client/builtin/TocList.vue
```diff
@@ -52,4 +52,7 @@ const classes = computed(() => {
 .slidev-layout .slidev-toc-item p {
   margin: 0;
 }
+.slidev-layout .slidev-toc-item div, .slidev-layout .slidev-toc-item div p {
+    display: initial;
```
antfu, line 56:
```suggestion
  display: initial;
```

PR #950, opened 2023-04-10
Add gap between columns in two-cols-* layouts

LostKobrakai:
This way the background of code boxes don't butt against each other.

### DIFF
--- a/packages/client/layouts/two-cols-header.vue
+++ b/packages/client/layouts/two-cols-header.vue
@@ -27,7 +27,7 @@ const props = defineProps({
     <div class="col-span-2">
       <slot />
     </div>
-    <div class="grid grid-cols-2">
+    <div class="grid grid-cols-2 gap-4">
       <div class="col-left" :class="props.class">
         <slot name="left" />
       </div>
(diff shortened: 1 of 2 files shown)


antfu:
I am not very sure. I feel it should leave to the users to decide.

LostKobrakai:
That arguments kinda works in both ways. The user can customize the default by adding or removing any gap. I however think having a gap is the better default over not having a gap.

LostKobrakai:
You can hack one in easier (margin on both columns), but adding a gap using the proper css (where it's one space properly aligned) takes the same amount of work of a custom layout as far as I see.

YuanYuYuan:
I second this. It would be great to have an appropriate gap by default.

maurer2:
Hello, I solved it by creating a custom layout `two-cols-with-gap.vue` in the layouts folder:

```vue
<script setup lang="ts">
const props = defineProps({
  class: {
    type: String,
  },
})
</script>

<template>
  <div class="slidev-layout w-full h-full">
    <div class="grid grid-cols-2 gap-4">
      <div class="col-left" :class="props.class">
        <slot name="left" />
      </div>
      <div class="col-right" :class="props.class">
        <slot name="right" />
      </div>
    </div>
  </div>
</template>
```

The new layout can then be applied in a slide like so:

```md
---
layout: two-cols-with-gap
hideInToc: true
---
```

Cheers

PR #960, opened 2023-04-18
fix: allow vite config override

Dtdhamu:
Hai


Dtdhamu (commented):
e17a433778d6dfdb2f7af445183fe566407055da

PR #970, opened 2023-04-20
Add option for exporting without `end` slide

pascalheidmann:


antfu:
The "end" slide should never be exported, if it does, it sounds like a bug we should fix

pascalheidmann:
If you only have the implicit end slide it isn't exported. But if you have a slide with `layout: end` it will be exported. 

Depending on your usecase this explicit end slide makes sense to export (eg you want your regular slides exported as pdf fallback) but for giving out your slides there is no need for it.

antfu:
I feel this is a bit nitch. Maybe instead of yet another cli flag, we could introduce a frontmatter for each slide to be more flexible:

Maybe:

```ts
---
hidden: export
---
```

PR #973, opened 2023-04-22
fix: remove rounded class to fix blurry text

a1ooha:
hover interface with rounded class

image

hover interface without rounded class

image


antfu:
Can you share more about your environment and why this would help?

antfu:
Closing due to inactivity. Please open a new issue referencing to this one if you can follow up with more information.

Issue #974, opened 2023-04-23
Disable default last page


HenriqueAurelio:
**Is your feature request related to a problem? Please describe.**
I'm trying to build an slide for apresentation but I'm not able to avoid the default end page.

**Describe the solution you'd like**
Is there a config to disable the default template or disable the last page?

**Describe alternatives you've considered**
Trying to use layout:end and build my last slide into it.

rightaway:
I also would like a setting to disable the end page. I already have created my own which is better than a blank slide with END.

@antfu Is there a workaround to remove it?

PR #1003, opened 2023-05-24
Option disable end slide (for #974)

Kazeheki:
With the new frontmatter-option 'includeDefaultEnd' the user can choose whether they want to:
- specifically include the last blank default end (true)
- specifically exclude the last blank default end (false)
- let slidev decide on the fact whether there is a user-added end-slide ('auto')

On export, the default blank page is always omitted but the user-added end-slide will be included.

PR for #974

### DIFF
--- a/packages/client/internals/PrintContainer.vue
+++ b/packages/client/internals/PrintContainer.vue
@@ -22,7 +22,9 @@ const scale = computed(() => {
 })
 
 // Remove the default "end" slide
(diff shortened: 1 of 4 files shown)


antfu:
I think maybe we should remove the end side as a whole (probably a bad decision on the early day), as users could always add their own end slide.

Kazeheki:
Were there any other reasons other than "would be nice to know the end" for the automatic end slide?
If not, I can create another PR to remove the automatic end slide if you want.
Do you have an Issue for that which I could link?

PR #1029, opened 2023-06-14
fix: prism highlighted line breaking

AlbertBrand:
The current implementation of line breaking of Prism-highlighted code is incorrect. It first splits code in separate lines and feed these to the highlight method. This does not work correctly for many kinds of code: JavaScript multiline strings and doc comments, XML attributes over multiple lines, etc.

As an alternative this PR sends the whole code to Prism and then splits line by line. A problem that then occurs: a newline can start when one or more tags are still open. In the PR I addressed this by parsing the HTML and closing/reopening tags for each line. I've searched the Prism API to see if there is some way around this, but haven't found it yet.


AlbertBrand:
@antfu Those lint errors were not introduced by my code it seems.

PR #1036, opened 2023-06-25
chore: upgrade unocss version

zyyv:
This is an upstream feature, but is a bug in slidev. As mentioned in https://github.com/unocss/unocss/pull/2770, uno will extract or transform my code, update the version for internal support.

e.g.
```
<!-- slide.md -->
```html
<div class="text-(red xl)">test</div>
\```
```

Will be transformed: 

```
<!-- slide.md -->
```html
<div class="text-red text-xl">test</div>
\```
```

but it's not what i want.


zyyv:
@antfu Sorry ping，could you please release a new version?

PR #1042, opened 2023-07-02
chore: update shiki dependency

IgnaceMaes:
https://github.com/shikijs/shiki/releases/tag/v0.14.3


antfu:
It's in the semver range, update your lock would upgrate shiki as well.

PR #1059, opened 2023-07-23
feat: option `editor` to disable built-in editor

### LINKED ISSUE #1048: How to disable Integrated Editor? The little audience ruined my ppt and lots of related files...
I've check all the guides [here](https://sli.dev/guide) and find no switches to cut off this feature...

How to disable it?

ps: I know build may probably solve this problem, but I didn't find a way to set password while hosting build version...

JoeyGrajciar:
proposed by user @15921483570 in linked issue

Closes #1048


antfu:
It's intentional. The "Dev" more is intended to be used for the author themselves, and should not expose to the audience directly. Generally, it's recommended to use a built SPA for the audience.

I think it would be great to:

- Disable the editor when a password is set to the presenter mode
- Have an option to disable it

JoeyGrajciar:
>  Generally, it's recommended to use a built SPA for the audience

Yes, that is true. There is #889 which describes issue with the SPA, where you can have no presenter mode or unprotected presenter mode. So instead @nejci is using password protected presenter mode in dev mode allowing outside world to use code mirror.

> Disable the editor when a password is set to the presenter mode
> Have an option to disable it

Sounds great. I will take a look on that.

There is also third option, implementing password protection for presenter mode in SPA. What do you think about it @antfu ?

antfu:
> implementing password protection for presenter mode in SPA

I don't mind having it. Tho maybe we need to make clear that it only visually blocks the view, but does not prevent malware actions.

/cc @tonai in case you have any other opinions

JoeyGrajciar:
Thanks, that might be for another PR.
Implemented your suggested changes to hide code editor:
- if password is set
- disabled explicitly via config

tonai:
> > implementing password protection for presenter mode in SPA
> 
> I don't mind having it. Tho maybe we need to make clear that it only visually blocks the view, but does not prevent malware actions.
> 
> /cc @tonai in case you have any other opinions

When you use static build, the presenter mode is not synced with other clients, so the need for a password is not obvious.
if we found a way to share it, that would then clearly become a problem, so this can still be a potential feature to add.
As you say, using a password in a client only app is clearly not a perfect protection, but I don't think we have another option.
And yes I also agree that this would be for another PR.

JoeyGrajciar:
> When you use static build, the presenter mode is not synced with other clients, so the need for a password is not obvious.

Okay, that is probably the missing piece of information for users. Documentation doesn't state mention that so that might be why presenter with password in dev is used.

Is there anything else that needs to be done in this PR?

JoeyGrajciar:
@antfu or @tonai anything else to be done in this PR?

packages/client/internals/NavControls.vue
```diff
@@ -118,7 +120,7 @@ if (__SLIDEV_FEATURE_DRAWINGS__)
           <carbon:user-speaker />
         </RouterLink>
 
-        <button v-if="__DEV__ && !isPresenter" class="slidev-icon-btn <md:hidden" @click="showEditor = !showEditor">
+        <button v-if="__DEV__ && !isPresenter && !passwordSet && configs.codeEditor" class="slidev-icon-btn <md:hidden" @click="showEditor = !showEditor">
```
tonai, line 123: Instead of only checking if password is set or not, maybe we can check if password is correct ?

PR #1060, opened 2023-07-26
feat: custom `backgroundSize` for image layouts (#798)

alpcoker:
fixes https://github.com/slidevjs/slidev/issues/798

Continue of the previous [PR](https://github.com/slidevjs/slidev/pull/862), accidentally deleted branch while rebasing.

With this additional prop to image layouts (image, image-left, image-right), we can now able give custom css [background-size](https://developer.mozilla.org/en-US/docs/Web/CSS/background-size). The default will remain `cover` 

My main motivation was to be able to use `contain`. Because I wasn't able to fit vertical photographs, let's cover all the different needs.  

Example usages:

```
---
layout: image
image: https://iili.io/HzdAhJV.png
backgroundSize: contain
---
```

```
---
layout: image-left
image: https://iili.io/HzdAhJV.png
backgroundSize: 20em 70%
---
```

On this PR, I also excluded 
- image-custom layout [from that discussion](https://github.com/slidevjs/slidev/pull/862#discussion_r1274314432)
- the demo starter commit [from that discussion](https://github.com/slidevjs/slidev/pull/862#discussion_r1274313100).

### DIFF
--- a/packages/client/layoutHelper.ts
+++ b/packages/client/layoutHelper.ts
@@ -9,7 +9,7 @@ export function resolveAssetUrl(url: string) {
   return url
 }
 
-export function handleBackground(background?: string, dim = false): CSSProperties {
+export function handleBackground(background?: string, dim = false, backgroundSize = 'cover'): CSSProperties {
   const isColor = background && (background[0] === '#' || background.startsWith('rgb'))
 
   const style = {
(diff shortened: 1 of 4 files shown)


packages/client/layouts/image-contain.vue
```diff
@@ -0,0 +1,18 @@
+<script setup lang="ts">
```
antfu, line 1: I think I am ok to have the rest. However, this file feels a bit too specific. Do you have any good use cases?

alpcoker: Thanks. For image-contain, I don't have any specific use cases, they were all alternatives. The best practical usage from my point of view would be from setting prop from existing layouts. 

I have dropped that commit & updated the PR.

PR #1072, opened 2023-08-05
feat: allow export presentation with notes on A4

oripka:
A new export for handouts. It creates an A4 page, the slide on the top and the speaker notes on the bottom.

The footer of the pages is customizable using custom components `handout-bottom.vue` similar to `global-bottom.vue` for slides.

It can optionally create cover pages by passing the `--cover` option, which then renders another custom component `handout-cover.vue`. Both these components need to be in your theme or project root.

Example usage:

```
cd demo/starter
node node_modules/@slidev/cli/bin/slidev.js --output handout export-handout --cover
```

Output is not a filename but used as a prefix here.

This creates the following files:

- `handout-slides.pdf` - slide PDF
- `handout-notes.pdf` - notes PDF
- `handout-handout-cover.pdf` - cover pages (optional)
- `handout.pdf` - final handout merged slides, notes and cover


DavidSichau:
Hi why was this closed. This would be an awesome feature.

PR #1073, opened 2023-08-06
feat: add lintSide preparser extension

oripka:
Add preparser extension `lintSlide` that allows for linting checks to be run on the parsed slide.


antfu:
Thank you for bringing this up, but as for Slidev we don't want to babysit on how users put their content. It's also easy to see when the content overflows. We probably not going to have it.

PR #1083, opened 2023-08-16
chore: add node compatibility badge

btea:


antfu:
Thank you, but I don't think it's necessary, to be honest.

PR #1143, opened 2023-10-06
fix(a11y): add missing alternative texts

### LINKED ISSUE #1142: a11y: Icon-buttons do not have an accessible name
**Describe the bug**
After running a lighthouse or axe accessibility report, it turned out, that all icon button do currently not have an accessible name.

**To Reproduce**
1. Start the slides
2. Generate an a11y report
3. check the rseults

for example: this buttons should have an accessible title (which can be visually hidden).
See this source for examples:
https://www.sarasoueidan.com/blog/accessible-icon-buttons/

 - OS: macOS
 - Browser: Chrome
 - Slidev version: 0.43.6

d-koppenhagen:
- provide `alt` attributes for `<image>` elements. Use`alt=""` for decorative images
- add `HiddenText` component for icon button usage. The component uses the WindiCSS class `src-only` to correctly hide elements visually but read texts out for screen readers

closes #1142

### DIFF
--- a/packages/client/internals/SlidesOverview.vue
+++ b/packages/client/internals/SlidesOverview.vue
@@ -9,6 +9,7 @@ import { getSlideClass } from '../utils'
 import SlideContainer from './SlideContainer.vue'
 import SlideWrapper from './SlideWrapper'
 import DrawingPreview from './DrawingPreview.vue'
+import HiddenText from './HiddenText.vue'
 
 const props = defineProps<{ modelValue: boolean }>()
 
@@ -157,6 +158,7 @@ watchEffect(() => {
     </div>
   </Transition>
   <button v-if="value" class="fixed text-2xl top-4 right-4 slidev-icon-btn text-gray-400" @click="close">
+    <HiddenText text="Close" />
     <carbon:close />
   </button>
 </template>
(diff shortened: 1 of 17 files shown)


packages/client/internals/SlidesOverview.vue
```diff
@@ -157,6 +158,7 @@ watchEffect(() => {
     </div>
   </Transition>
   <button v-if="value" class="fixed text-2xl top-4 right-4 slidev-icon-btn text-gray-400" @click="close">
+    <HiddenText text="Close" />
```
antfu, line 161: I wonder if in this case, would a `title` to the button would be more straightforward?

d-koppenhagen: I think the title attribute isn't the best way to make thinks more accessible (even if at the first look, I thought the same). See for example: https://www.a11yproject.com/posts/title-attributes/#:~:text=Short%20answer%3A%20Don't%20use,and%20possibly%20not%20even%20used.

It could be tackled with aria attributes as well, but there are also some older screenreaders and browser *caught* IE 😃 having problems with hidden elements.
The CSS trick is a pretty old but effective one which should work for almost every device and browser.

antfu:
Thanks!

PR #1153, opened 2023-10-13
refactor: remove using resolvePackageData from Vite

bluwy:
https://github.com/vitejs/vite/pull/14584 is removing the API for future perf optimizations and bug fixes. Since the API is not widely used, I've made a PR to refactor it away directly.

This uses [`vitefu`](https://github.com/svitejs/vitefu) which I maintain on the side and implements the same heuristic, and is also used by frameworks like Svelte, Solid, and Astro today.

(I've not tested this particular code, but I hope the CI covers it)


antfu:
I can't unsee it as Vite Fu 🤣

bluwy:
Now that you mention it 😄 If it helps unseeing it, it stands for Vite framework utilities.

Issue #854, opened 2023-01-30
VSCode - format slides.md - break the build


deleonio:
**Describe the bug**
If I format the markdown file - the build break.

**before**
```md
---
layout: center-layout
---
```

**after** formatting
```md
---

## layout: center-layout
```

**To Reproduce**
1. Run the `dev`
1. Go to `slides.md`
2. Make some slides with different layouts
3. Format the Markdown file `slides.md`
4. See the error

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: Chrome
 - Slidev version: 0.39

joe733:
This happens if you're using either [`Markdownlint`](https://github.com/DavidAnson/vscode-markdownlint) or [`Prettier`](https://github.com/prettier/prettier-vscode) as formatter. If you use [`MarkdownAllInOne`](https://github.com/yzhang-gh/vscode-markdown) instead, that particular change does not happen.

Not so cool suggestion: Adding a formatter to [`slidev-vscode`](https://github.com/slidevjs/slidev-vscode)

maxandron:
A cool suggestion would be to add a prettier parser :D

emersonbottero:
I just add an the prettier config file in the root disabling it for .md

wycats:
I think this should be reopened. It's probably my biggest pain point with slidev atm.

jakebailey:
I've so far gotten through this with `<!-- prettier-ignore-start -->` or `<!-- dprint-ignore-start -->`, like:

```
<!-- prettier-ignore-start -->

---
layout: fact
---

<!-- prettier-ignore-end -->
```

But, another way to do this would be great.

maxandron:
@jakebailey this is very cool! I didn't know this option existed. Thanks!

Now creating a workaround is quite simple: <https://gist.github.com/maxandron/ce7ca54f2e808dadd209e34ebe751f92>

jakebailey:
That works for those who only run prettier separately; I format in my editor, so, I think an alternative way to write these blocks would be still great.

Maybe:

```md

---
title: the first one is okay
---

# Blah

---

<v-config>
layout: fact
</v-config>

# This is a fact!

```

Or something. Unfortunate that this repo uses the stale bot, though; this is the second filing of this 😢

CodyBontecou:
[...]

PR #1186, opened 2023-11-09
feat: the first yaml code block as frontmatter

mskelton:
Fixes https://github.com/slidevjs/slidev/issues/854

This adds support for configuring slide frontmatter using an alternative `<v-config>` syntax to make it easier for Prettier users to change settings per slide rather than the typical frontmatter approach.

```md
---
layout: cover
---
```

Can now be written like this:

```md
---

<v-config>
layout: cover
</v-config>
```

This doesn't change the existing method of using per-slide frontmatter, that's still supported without any changes.


antfu:
Maybe

~~~md
---

```yml
layout: cover
```
~~~

So it works better with syntax highlight

mskelton:
Great idea. Updated the PR to detect if it's the first code block immediately after the slide break and is `yml` or `yaml`.

Issue #1190, opened 2023-11-13
v-clicks not working in combination with custom components and <slot>


patrickpichler:
**Describe the bug**
`<v-clicks>` is not working when used in a custom component and passed children via `<slot>`. It simply shows
all the children, but you cannot click through it.

**To Reproduce**
1. Clone https://github.com/patrickpichler/slidev-components-sample/
2. Checkout commit `297962cd93be5f119164d26416cc5d9331fcfdfc`
3. Run `npx slidev`

**Desktop (please complete the following information):**
 - OS: MacOS
 - Browser: Firefox + Chrome
 - Slidev version: 0.43.15

twitwi:
self-note (to detect slots, if nothing better):
`(listTags.includes(i.type as string) || typeof i.type === 'symbol')`
`(listTags.includes(defaults[0].type as string) || typeof defaults[0].type === 'symbol')`

PR #1191, opened 2023-11-14
fix: handle slot in v-clicks

twitwi:
fixes https://github.com/slidevjs/slidev/issues/1190

This is a fix for the issue, i.e. it makes it possible to have a component defined somewhat like this:

```html
<template>
    <v-clicks><slot></v-clicks>
</template>
```

NB: slots are detected using typeof i.type === 'symbol', I'm not sure if there is a more robust way (like importing the symbol from vue implementation?)

NB: another case, that is more difficult to handle is the one where we have a component that renders e.g. as an `ul` and that is used in a nested list, this is currently (still) missed by v-clicks as its logic works on the un-rendered component.


antfu:
Could you add a simple e2e tests to avoid future regressions? Thanks

twitwi:
Ok, here they are.
That was a good opportunity to discover cypress.

PR #1196, opened 2023-11-21
Add --browser option to export

### LINKED ISSUE #1195: Export using Firefox (or any browser engine Playwright supports)
I usually present my slides with Firefox because of some i18n-related typography things Chrome does not support, but it is not an option for `slidev export`; exported PDF has unwanted layout.

Since Playwright is shipped with multiple engines, is it possible to have an option to choose which one to render?

yheuhtozr:
Closes #1195.

Not tested yet because I was not able to correctly build this package locally.

### DIFF
--- a/packages/slidev/node/export.ts
+++ b/packages/slidev/node/export.ts
@@ -32,6 +32,7 @@ export interface ExportOptions {
    * @default false
    */
   perSlide?: boolean
+  browser?: 'chromium' | 'firefox' | 'webkit'
 }
 
 function addToTree(tree: TocItem[], info: SlideInfo, slideIndexes: Record<number, number>, level = 1) {
@@ -65,6 +66,7 @@ export interface ExportNotesOptions {
   base?: string
   output?: string
   timeout?: number
+  browser?: 'chromium' | 'firefox' | 'webkit'
 }
 
 function createSlidevProgress(indeterminate = false) {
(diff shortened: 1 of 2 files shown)


packages/slidev/node/export.ts
```diff
@@ -108,12 +110,18 @@ export async function exportNotes({
...
+  if (!browserOption) {
+    throw new Error("Available Playwright browsers are: chromium, firefox, webkit")
+  }
+  const browserPackage = `playwright-${browserOption}`;
+  if (!packageExists(browserPackage))
+    throw new Error('The exporting for Slidev is powered by Playwright, please install it via `npm i -D ' + browserPackage + '`')
 
-  const { chromium } = await import('playwright-chromium')
-  const browser = await chromium.launch()
+  const { playwright } = await import(browserPackage)
```
yjl9903, line 123: Does this works fine?

I think the following seems better. Make bundler more friendly knows which is imported here, instead of a dynamic string, even if you have checked before.

```js
function loadBrowser(browserPackage) {
  if (browserPackage == 'chromium') {
    return import('playwright-chromium')
  } else ...
}
```

KermanX:
Closing this, as pointed out by https://github.com/slidevjs/slidev/pull/1196#issuecomment-1979681462

PR #1209, opened 2023-12-04
fix: allow more characters for slot names

### LINKED ISSUE #1208: Special characters in slot names
Certain characters such as `-[].,` are not allowed in the shorthand slot syntax (`::slot-name::`), while they are allowed with the `<template #slot-name />` syntax.

**To Reproduce**
1. Create a Component with a named slot containing the characters above:
   ```vue
   <template>
     <div>
       <slot />
       <div class="slot1">
         <slot name="slot.1" />
       </div>
       <div class="slot2">
         <slot name="slot-[2]" />
       </div>
     </div>
   </template>
    
   <style scoped>
   .slot1 {
     background: pink;
   }
    
   .slot2 {
     background: slateblue;
   }
   </style>
   ```
3. The syntactic sugar syntax does not work:
   ```md
   # Slide
   <Component>
   
   ::slot.1::
   Does not work

   <template #slot-[2]>

   Does work

   </template>  
   </Component>
   ```

 - OS: Ubuntu 22.04
 - Browser: Google Chrome Version 119.0.6045.123 (Official Build) (64-bit)
 - Slidev version: main branch

0phoff:
Closes #1208.

Switched `\w` to `\S` in the regexp for matching the `::slot-name::` syntax.
`\S` matches any non white-space character, not sure if it is the right group or if we need some other custom matching list.
I am not familiar enough with Vue to know which characters are allowed for slot names.

### DIFF
--- a/packages/slidev/node/plugins/markdown.ts
+++ b/packages/slidev/node/plugins/markdown.ts
@@ -144,7 +144,7 @@ export function transformSlotSugar(md: string) {
     if (isLineInsideCodeblocks(idx))
       return
 
-    const match = line.trimEnd().match(/^::\s*(\S+)\s*::$/)
+    const match = line.trimEnd().match(/^::\s*([\w\.\-\:]+)\s*::$/)
     if (match) {
       lines[idx] = `${prevSlot ? '\n\n</template>\n' : '\n'}<template v-slot:${match[1]}="slotProps">\n`
       prevSlot = true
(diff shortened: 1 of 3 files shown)


0phoff:
> I guess we could have `[\w._-]` instead of `\S`. Not all chars are valid HTML attr.

I imagined Vue/Vite would crash on invalid slot-names, but using a stricter set in the parser is probably nicer. I think "_" is already included in `\w`, but otherwise this seems correct IMO.

I'll make the change!

0phoff:
Added stricter slotname matching.

One minor remark is that the characters `[]` are not added.
From my testing this seems to be valid for Vue slot names, but as they are invalid characters for HTML attrs, I left them out.

Personally, I have no need for them and only wanted `-` to be added, so I'm fine with leaving them out! :innocent:

PR #1222, opened 2023-12-20
feat: support using external file as code snippet, fix #1193

### LINKED ISSUE #1193: Proposals: include code snippet from external source file
Usually to make sure the code on the slide is correct you need to compile and test it.
Then you copy it into the slide.
But code can change. Now you need to copy & paste it again.
It would be nice to just include the required lines from the external source file automatically without copy/paste.

````
``` cpp [./src/list.hpp: 10 | 20] {all | 10 | 14-16}

```
````

will read lines 10 to 20 from /src/list.hpp file and display that as a snippet.

Alternative:
Use a component:
```html
<SourceCode file="./src/list.hpp" from="10" to="20"  anim="all | 10 | 14-16" />
```

KermanX:
For #1193.

Support using external files as code snippets like:

```md
<<< @/path/snippet.js#region ts {1|2}{lines:true}
```

Then it will display the "region" region in "/path/snippet.js", and language set to `ts` instead of `js`, with metadata `{1|2}{lines:true}`

Syntax is mostly the same as [vitepress#import-code-snippets](https://vitepress.dev/guide/markdown#import-code-snippets). See [this comment](https://github.com/slidevjs/slidev/pull/1222#issuecomment-1896124801).

(edited after antfu's comment)


packages/parser/src/fs.ts
```diff
@@ -34,6 +34,27 @@ export async function load(filepath: string, themeMeta?: SlidevThemeMeta, conten
     const baseSlide = data.slides[iSlide]
     if (!baseSlide.frontmatter.src) {
       iSlide++
+
+      baseSlide.content = baseSlide.content.replaceAll(
+        /^```(\w+?)\s*\[([\s\S]+?)\]([\s\S]*?)\n+^```/mg,
+        (full, lang = '', external: string, rest: string) => {
+          const [externalPath, externalRangeStr] = external.split(':').map(i => i.trim())
+          const sourcePath = resolve(dir, 'snippets', externalPath)
```
KermanX, line 42: I am not sure should the snippets be put in `${dir}/snippets/${snippetName}` or somewhere else, or should use `userRoot` instead of `dir`.

KermanX:
Is anyone here?

packages/parser/src/fs.ts
yjl9903: I think the snippet path should be resolved relative to the current slide markdown file. This seems more make sense?

packages/parser/src/fs.ts
```diff
@@ -34,6 +34,27 @@ export async function load(filepath: string, themeMeta?: SlidevThemeMeta, conten
...
+          entries.add(sourcePath)
+          let source: string
+          if (!existsSync(sourcePath)) {
+            lang = 'plaintext'
+            source = `File not found: ${sourcePath}`
+          }
+          source = readFileSync(sourcePath, 'utf-8')
```
yjl9903, line 49: It may throw some other errors here. Catch them here?

packages/parser/src/fs.ts
KermanX: I had a look at L76-82 of this file, maybe this is a better choice?

```ts
let sourcePath
if (external.startsWith('/'))
  sourcePath = resolve(dir, external.substring(1))
else if (baseSlide.source?.filepath)
  sourcePath = resolve(dirname(baseSlide.source.filepath), external)
else
  sourcePath = resolve(dir, external)
```

(The same rule as the `src` in frontmatter)

packages/parser/src/fs.ts
KermanX: Resolved in 30e77d5

antfu:
Thanks for working on it! I indeed love that feature!

However, I am a bit concerned that we are introducing too much syntax in the code snippet meta string, and might make it confusing in the long term. Maybe we could result VitePress' syntax to do so, it could also lowering the learning cost as the syntax would word across more ecosystem:
https://vitepress.dev/guide/markdown#import-code-snippets
https://vitepress.dev/guide/markdown#markdown-file-inclusion

We may directly use the implementation they made

KermanX:
I have rewritten the code to VitePress-like syntax.

There are some differences between this PR and VitePress:

- VitePress has a `title` field
- VitePress has only one `meta`

So the syntax introduced by this PR is:

```md
<<< @/path/snippet.js
<<< @/path/snippet.js#region ts {1|2}{lines:true}
// VitePress: <<< @/path/snippet.js#region {1,2 ts} [title]
```

The second one's language will be `ts` instead of `js`.

HMR is tested manually and tests are updated. The usage of `#region` is the same as in VitePress.

Markdown inclusion is not included in this PR yet. Maybe in another PR it can be implemented.

packages/slidev/node/plugins/transformSnippet.ts
```diff
@@ -0,0 +1,128 @@
+import path from 'node:path'
```
antfu, line 1: Here in this file, can we have a reference to VitePress' implementation source code? Put some credits while make easy to track with the upstream updates.

antfu (approved):

levchak0910:
Hi @KermanX, thanks for implementing it )

Regions are great way to reuse code

> So the syntax introduced by this PR is:
> 
> ```
> <<< @/path/snippet.js
> <<< @/path/snippet.js#region ts {1|2}{lines:true}
> // ❌ VitePress: <<< @/path/snippet.js#region {1,2 ts} [title]
> ```

Could you please mention the difference between syntax highlighting in slidev and vitepress in the documentation? Currently this knowledge available only in this PR.

PR #1228, opened 2023-12-25
fix: Highlighting latex equations when using \begin{align*}

John98Zakaria:
Consider the following `slides.md` file

```
---
layout: default
---
# Slide

$$ {1|2|3|4}

\begin{align*}

R=& (111....)_2 + (111....)_2 \\
=& \sum_{n=0}^{n-1} 2^n + \sum_{n=0}^{n-1} 2^n \\

=& \: 2*{\sum_{n=0}^{n-1} 2^n} \\

=& \: 2*(2^{n} - 1)

\end{align*}

$$

```

I would expect the [highlighting to work for LaTeX ](https://sli.dev/guide/syntax.html#latex-line-highlighting).
However it doesn't.

This generalizes the logic of finding the individual rows by iterating over all .col-align-XXX classes that contain the individual equation rows.

PR #1255, opened 2024-01-21
fix: Latex line highlighting for vector equations

John98Zakaria:
PR: https://github.com/slidevjs/slidev/pull/1228

Introduced a regression in which vector symbols get classified as a latex line, causing the line highlighting to break.

By adding [:scope](https://developer.mozilla.org/en-US/docs/Web/CSS/:scope) to the css selector only the direct descendants on the parent will be selected, thereby skipping child elements (vectors) that have the same css structure

```
---
layout: default
---
# Slide

$$ {0|1|2|3|all}
\begin{array}{c}

\nabla \times \vec{\mathbf{B}} -\, \frac1c\, \frac{\partial\vec{\mathbf{E}}}{\partial t} &
= \frac{4\pi}{c}\vec{\mathbf{j}} \nabla \cdot \vec{\mathbf{E}} & = 4 \pi \rho \\

\nabla \times \vec{\mathbf{E}}\, +\, \frac1c\, \frac{\partial\vec{\mathbf{B}}}{\partial t} & = \vec{\mathbf{0}}\\

\nabla \cdot \vec{\mathbf{B}} & = 0 

\end{array}
$$
```

Before (clicks=1)

image

After the fix

image

### DIFF
--- a/packages/client/builtin/KaTexBlockWrapper.vue
+++ b/packages/client/builtin/KaTexBlockWrapper.vue
@@ -70,7 +70,7 @@ onMounted(() => {
 
     // For each row we extract the individual equation rows
     const equationRowsOfEachParent = Array.from(equationParents)
-      .map(item => Array.from(item.querySelectorAll('.vlist-t > .vlist-r > .vlist > span > .mord')))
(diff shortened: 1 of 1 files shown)


John98Zakaria:
I really want to apologize for introducing this bug especially that this equation is included as an example for the line highlighting feature

PR #1265, opened 2024-02-02
feat!: migrate to shiki v1

antfu:
This PR removes `highlighter: shikiji` support. Use `highlighter: shiki` instead. Breaking changes between shiki v0.x to shiki v1 applies. For example, if you are using `loadTheme` in shiki setup, change it to manual fs reads instead.

This PR also improves the TwoSlash support by using [floating-vue](https://floating-vue.starpad.dev/)


John98Zakaria:
I was using shikji as a highlighter as it had a nicer theme.
Do you happen to know how I can get the theme back in Shiki ?

Before (transparent background)
image

After
image

PR #1273, opened 2024-02-05
feat: add runnable monaco editor (#219)

### LINKED ISSUE #219: notebook cell element
Hi, first I have to say this is a very nice work! thanks for making this.

In most of the developer presentations, I believe it happens that one needs to demonstrate some code,
which is a cool feature for jupyter notebooks with [revealjs](https://revealjs.com/) via the [RISE plugin](https://rise.readthedocs.io/en/stable/).
This makes demonstrating a piece of code reactively a lot easier, so that one do not need to copy the code inside the slide to a terminal REPL window
manually. So I believe it would be nice if slidev could have this.

However, the RISE plugin has a very limited layout one could choose and do not have other nice features slidev has.

I'm thus thinking if there could be a slidev component that allows one to insert a notebook cell anywhere they like, e.g
for the compiled HTML element there could be a class of cell denote which `div` is the cell,

```html
<div class="notebook-cell"></div>
```

this `div` should save the output produced by excuting the inner code cell.
and for markdown syntax, it could be a special code block like how monaco works

    ```js {cell}
    const count = ref(1)
    const plusOne = computed(() => count.value + 1)
    
    console.log(plusOne.value) // 2
    
    plusOne.value++ // error
    ```

potentially this could connect to a jupyter notebook kernel, or just run on its own communication
kernel and in the export mode, we can save the results as HTMLs and force the cell to be converted
to a static code block.

I know there might be some related work done in the Julia-based notebook Pluto: https://github.com/fonsp/Pluto.jl
but it's less general for other languages since it doesn't aim to support other languages, and does not focus on writing
slides.

KermanX:
### Overview

resolves #219

This PR introduces the "runnable" Monaco editor:

**Grammar**:
~~~
```ts {monaco-run}
console.log('Hello' as string, [1, 'text', new Error('message'), () => {}])
```
~~~

**Example**:
image

https://github.com/slidevjs/slidev/assets/11247099/0c6ce681-80d3-4555-93bf-9288ee533462

### Options

- `autorun`: `boolean | 'once'` 
   Default: `true`
   Run the code automatically after input
- `outputHeight`: `string | undefined`
   Default: `undefined`
- `highlightOutput`: `boolean`
- `runnerOptions`: `Record<string, unknown>`
  Will be passed to the runner as-is.

### Notes

- Use `setup/runners.ts` to provide custom runners. You can highlight code and call other runners in your runner.
- Code will not run in the presenter preview or in the overview.
- Code will always be autorun in print mode.
- Using `new Function()` to run the code.
- Cannot use import statements in the code. Should use dynamic import instead.

### Todos

- [x] Print mode
- [ ] Not sure whether "monaco-runnable" is the best name? (`{monaco-run}`?)
- [ ] Any better way to import code, and how to import local code?


packages/client/builtin/Monaco.vue
```diff
@@ -141,10 +155,131 @@ onMounted(async () => {
...
 </script>
 
 <template>
-  <div ref="outer" class="slidev-monaco-container" :style="{ height }">
-    <div ref="container" class="absolute inset-0.5" />
+  <div class="relative" :data-waitfor="runnable ? '.output' : undefined">
+    <div ref="outer" class="slidev-monaco-container" :style="{ height }">
+      <div ref="container" class="absolute inset-0.5" />
+    </div>
+    <template v-if="runnable">
```
antfu, line 215: Could we make runnable a separate component to make this easier?

packages/client/internals/MonacoOutput.vue
```diff
@@ -0,0 +1,143 @@
...
+    <div v-else-if="output === '_empty'" class="text-sm text-center opacity-50">
+      Click the play button to run the code
+    </div>
+    <div v-else-if="output === '_running'" class="text-sm text-center opacity-50 running">
+      Running...
+    </div>
+    <template v-else>
+      <div class="mb-1 -mt-1 text-xs font-bold text-primary">
+        OUTPUT
+      </div>
```
antfu, line 120: Made the change to use Vue to render instead of assign everything in HTML, this should give us better control for the styling, and also interactivity if needed in the future.

I also think we could have a `raw` module or so, where we prints the output plain text (without `LOG` indicator and so), for example I'd expect `console.log('Hello World')` should just printing plain `Hello World` instead of `LOG: "Hello World"`.

packages/client/internals/MonacoOutput.vue
```diff
@@ -0,0 +1,143 @@
...
+    </template>
+  </div>
+  <div v-if="code.trim()" class="absolute right-1 top-1 max-h-full flex gap-1">
+    <IconButton
+      class="w-8 h-8 max-h-full flex justify-center items-center"
+      :title="props.autorun ? 'Rerun' : 'Run code'"
+      @click="run"
+    >
+      <carbon:renew v-if="props.autorun" />
+      <carbon:play-filled-alt v-else />
```
antfu, line 130: I found the "Refresh" button a bit confusing, maybe we could always show the play button instead?

```suggestion
      <carbon:play-filled-alt />
```

packages/client/internals/MonacoOutput.vue
```diff
@@ -0,0 +1,143 @@
...
+    </IconButton>
+  </div>
+</template>
+
+<style scoped lang="postcss">
+.output :deep(.log-type) {
+  @apply font-bold op-70;
```
antfu, line 137: Removed scoped, instead we should prefix every class with `slidev-`. This would make it easier for ppl to override the style. Scoped classes are very hard to work with from the outside.

packages/client/internals/MonacoOutput.vue
```diff
@@ -0,0 +1,143 @@
...
+      `<span class="decorator">[</span>`,
+      `<span class="log-type ${type}">${type}</span>`,
+      `<span class="decorator">]:&nbsp;</span>`,
+      `<div class="select-text">`,
+      content.map(highlight).join('<span class="decorator">, </span>'),
+      `</div>`,
+      `</div>`,
+    ],
+  ).join('')
+
```
antfu, line 82: Would be nice to render Errors as well, in some cases we might want to present certain errors in the slides.

packages/client/internals/MonacoOutput.vue
```diff
@@ -0,0 +1,167 @@
...
+          : themes.light || 'vitesse-light',
+    })
+  }
+
+  const { transpile } = (tsModule ??= await import('typescript'))
+  const js = props.lang === 'typescript'
+    ? transpile(code.value, {
+      module: tsModule.ModuleKind.ESNext,
+    })
+    : code.value
```
antfu, line 65: We might want to throw a warning for non ts/js lang

KermanX: Added a `setup/runners.ts` to register custom runners.

packages/client/internals/MonacoOutput.vue
KermanX: Inspired by your advice, I am now wondering if `LOG` indicator is actualy needed. Most of the code people want to run in slides may only use `console.log`.

KermanX: Also, now `console.log('Hello')` will just output `Hello`, while `console.log(['Hello'])` will output `["Hello"]`.

KermanX: Removed log type indicator and `rawMode` option in b3792508929a26b32540fb0cdb7e93280ede1db0

packages/types/src/runner.ts
```diff
@@ -0,0 +1,30 @@
+import type { CodeToHastOptions } from 'shiki'
+
+export interface RunnerContext {
```
antfu, line 3:
```suggestion
export interface CodeRunnerContext {
```

Let's prefix all `Runner` with `CodeRunner` to be more explict.

antfu:
Looks very nice! I like it!

We could have some docs for it, and maybe add one new `demo` to experiments with the runner step? (otherwise, we have no tests for it)

PR #1279, opened 2024-02-08
fix!: rework the clicks system

### LINKED ISSUE #805: 🐞 v-click sometimes break code highlighting
**Describe the bug**
Stackblitz example: https://stackblitz.com/edit/slidev-yub2x6?file=slides.md

If we have some complex structure in our md then code highlighting gets broken.

**To Reproduce**
1. Go to reproduced [link](https://stackblitz.com/edit/slidev-yub2x6?file=slides.md)
2. Go to second slide and reach to click 5
3. Now on step 5 it should highlight `const b = 2` but it doesn't
4. Next on step 6 it should highlight `const c = 3` but it doesn't. Instead it highlights `const b = 2` 🤔 
5. Now on step 7 it is removing code as expected but we didn't get highlighted line for `const c = 3`

 - OS: Apple M1
 - Browser: Version 108.0.5359.124 (Official Build) (arm64)
 - Slidev version: v0.38.1

---

<details>
<summary>My content to test further</summary>

```md
<div class="grid grid-cols-2 grid-gap-4 h-full">

<div>

  ## First Visit
  
  <ul class="!list-decimal text-10px [&_p]-m-0">
    <li v-click="1">First vue app gets loaded</li>
    <li v-click="2" v-md>Along with other plugins, [CASL](https://casl.js.org/v6/en/package/casl-vue) (Access Control library) gets initialized</li>
    <li v-click="3" v-md>CASL initializes user ability. It will try to fetch it from localStorage if exist or will use from `initialAbility` const defined in the same file. _(File: `@/plugins/casl/ability.ts`)_</li>
    <li v-click="7" v-md>All plugins are loaded. Vue router starts navigation.</li>
    <li v-click="8" v-md>Router's [beforeEach](https://router.vuejs.org/guide/advanced/navigation-guards.html#global-before-guards) guard triggers</li>
    <li v-click="9" v-md>We checks if users is logged in via `isUserLoggedIn` method and assign it to `isLoggedIn` const. `isUserLoggedIn` method is in `@/router/utils.ts`. If `userData` and `accessToken` both keys are exist in localStorage then it returns `true` indicating user is logged in or else return `false` indicating user isn't logged in.</li>
    <li v-click="10">
      <p v-md>If user can navigate to requested route _(calls `canNavigate` method in `@/@layouts/plugins/casl.ts`)_</p>
      <ol>
        <li v-click="11">
          <p v-md>**If True,** check if user is logged in & has `redirectIfLoggedIn` route meta set to `true`.</p>
          <ol>
            <li v-click="12" v-md>**If True,** redirect user to index route. This is useful for pages like login where we don't want user to visit login page if he/she is already logged in.</li>
            <li v-click="13" v-md>**If False,** redirect user to requested route. It's normal flow, if user wants to navigate to dashboard let he/she navigate.</li>
          </ol>
        </li>
        <li v-click="14">
          <p v-md>**If False,** check if user is logged in</p>
          <ol>
            <li v-click="15" v-md>**If True,** redirect user to **not authorized** page.</li>
            <li v-click="16" v-md>**If False,** redirect user to login page</li>
          </ol>
        </li>
      </ol>
    </li>
  </ul>
</div>

  <div class="[&_p]-m-0 flex items-center justify-center">

<div v-click-hide="2" v-click="1">

```html
<!-- /index.html -->
<script type="module" src="/src/main.ts"></script>
```

</div>

<div v-click-hide="3" v-click="2">

```ts{2,13-15}
// File: main.ts
import ability from '@/plugins/casl/ability'

// other imports

// Create vue app
const app = createApp(App)

// Use plugins

// other plugins

app.use(abilitiesPlugin, ability, {
  useGlobalProperties: true,
})

// Mount vue app
app.mount('#app')
```

</div>

<div v-click-hide="6" v-click="3" style="--slidev-code-font-size:8px" class="relative">

```ts{15|16-18|4-9,20}
import { Ability } from '@casl/ability'
import type { UserAbility } from './AppAbility'

export const initialAbility: UserAbility[] = [
  {
    action: 'read',
    subject: 'Auth',
  },
]

//  Read ability from localStorage
// 👉 Handles auto fetching previous abilities if already logged in user
// ℹ️ You can update this if you store user abilities to more secure place
// ❗ Anyone can update localStorage so be careful and please update this
const stringifiedUserAbilities = localStorage.getItem('userAbilities')
const existingAbility = stringifiedUserAbilities
  ? JSON.parse(stringifiedUserAbilities)
  : null

export default new Ability(existingAbility || initialAbility)
```

</div>

  </div>
</div>
```

</details>

### LINKED ISSUE #863: Hot relading breaks `v-clicks` CSS classes
**Describe the bug**

When hot-reloading a slide with a `<v-clicks>` component, the `.slidev-vclick-current` class breaks, instead `.slidev-vclick-prior` is used until you reload the page manually.

For some slides I am setting `display: none` for `.slidev-vclick-prior` to switch between different animation states (useful for LaTeX animations until #266 is implemented), so this bug means that I can't see the changes after hot reloading, because everything inside `v-clicks` is invisible until I reload the page manually. Especially when debugging errors in LaTeX formulas this can limit productivity.

**To Reproduce**

1. Create a slide with this content:

```html
<style>
  .slidev-vclick-prior {
    color: lightblue;
  }

  .slidev-vclick-current {
    color: lightgreen;
  }
</style>

<v-clicks>

1. One
2. Two
3. Three

</v-clicks>
```

3. Launch it with `npx slidev` and navigate to <http://localhost:3030/1?clicks=2> (i.e. 2 clicks in).
4. Observe that item 1 is blue, item 2 is green and item 3 is hidden.
5. Using the inspect tools of your browser, observe that item 1 has class `slidev-vclick-prior` and item 2 has class `slidev-vclick-current`.
6. Apply a change to the slide, e.g. changing "Three" to "Thre" to trigger hot reloading.
7. Observe that both item 1 and 2 are blue.
8. Using the inspect tools of your browser, observe that item 1 and 2 have class `slidev-vclick-prior`.
9. Refresh the page.
10. Observe that the intended result (step 4 and 5) is visible again.
11. Change the style of `.slidev-vclick-prior` to `visibility: hidden;`.
12. Trigger hot reloading again.
13. Observe that you can't see anything anymore, because of the incorrectly applied CSS class.

 - OS: Linux (Pop!_OS)
 - Browser: Firefox 109.0 (64-bit)
 - Slidev version: `@slidev/cli": "^0.39"`

### LINKED ISSUE #1136: Code blocks hightlighting steps & `v-click` don't work well together
**Describe the bug**
When using code blocks with hightlighting steps and `v-click` (only with [custom click count](Code blocks hightlighting steps & `v-click`) apparently, but has TBC), the behaviour of the animations is a bit messy (repeated clicks, some parts do not appear).

**To Reproduce**
1. Go to https://stackblitz.com/edit/stackblitz-starters-gjnzi8?file=slides.md%3AL17-L21
2. Navigate through slides, they are self-explanatory (see also the highlighted lines in `slides.md` (L.17 to 21) which are never shown up)

**Expected**
On the second page, the following click behaviour is expected :
- `clicks=0`: First line highlighted
- `clicks=1`: Second line highlighted
- `clicks=2`: Second line highlighted (or all, but only second seems more appropriate)
- `clicks=3`: Same as `clicks=2` + the `div` text shows up

 - OS: -
 - Browser: -
 - Slidev version: 0.43.5
 -

### LINKED ISSUE #1231: Synchronisation between Highlight and v-click
Hello 
I start a presentation for my study, and I need your help. 

**Describe the bug**
The v-click and Highlight is not synchronise. I D'ont know if is my config (basic) or If I do something wrong but the click for the Highlight start after all the other click. So It's no synchronise. 
I can fix it with defined the good click to my text to correspond but it's not a solution... I have 40 slides ^^

I try to put without div, with. 
When I have no texte at left there is no conflict. I think it's a story of priority in the source code.. 

** The code **
--- 
layout: two-cols
layoutClass: items-center
hideInToc: true
---

<div v-click="8" text-xl>Gestion du formulaire :
<li v-click="8"><strong>currentTarget</strong> : Récupère l'élément du formulaire ciblé</li>
<li v-click="9"><strong>formData</strong> : Creer une nouvel instance de FormData basé sur le formulaire</li>
<li v-click="10"><strong>formObj</strong> : Transforme FormData en object JavaScript</li>
<li v-click="11"><strong>append</strong> : Ajoute des dates au formulaire sous le format necessaire au back-end</li>
<li v-click="12"><strong>dispatch dans Redux</strong> : Envoyer Actions, Mettre à jour État.</li>
<li v-click="13"><strong>axiosInstance</strong> : Requêtes HTTP, Interactions Serveur.</li>
</div>

::right::

```ts {all|1-3|4-5|5-6|8-10|17-|11-17} {at:0}
const handleSubmitAddUserChoice = (event: FormEvent<HTMLFormElement>) => {
  event.preventDefault();
  const form = event.currentTarget;

  const formData = new FormData(form);
  const formObj = Object.fromEntries(formData);

  formData.append('startDate', startDate);
  formData.append('endDate', endDate);

  if (startDate && endDate) {
      axiosInstance
        .post('userchoice', formObj, { withCredentials: false,})
        .then((response) => {
          return JSON.parse(JSON.stringify(response.data));
        });
      dispatch(
        showNotification({
          message: 'Votre date a bien été ajoutée', type: NotificationType.Success,})
      );
      setTimeout(() => {
        window.location.reload();
      }, 1000);
  } else {
      dispatch(
        showNotification({
          message: "Oops quelque chose s'est mal passé",
          type: NotificationType.Error, 
  }));}};
```

 - OS: Mac 14.1.00
 - Browser: I try on Edge / Safari / Chrome and Mozilla 
 - Slidev version: 0.46.1

### LINKED ISSUE #1246: Highlighting two blocks of code in the same time doesn't work
**Describe the bug**
I can't make the `{at:0}` feature work when highlighting lines of code blocks. I have tried to use the solution of [this discussion](https://github.com/slidevjs/slidev/discussions/431#discussioncomment-2625444) to highlight the lines of two blocks of code at the same time but it doesn't work (highlighting is done in sequence as if the `{at:0}` was ignored).  

**To Reproduce**
1. Initialize a new presentation with `npm init slidev`. Replace the content of the `slides.md` with this (remove the `\` before the ` ``` `):
```
\```js{all|1|2|all}
const a = "Highlight me first!"
const b = "then me!"
\```

\```js{all|1|2|all} {at:0}
const thisVarChanged = "Highlight me first!"
const eyo = "then me!"
\```

---

\```js{all|1|2|all}
const a = "Highlight me first!"
const b = "then me!"
\```

<v-clicks at="0">

* one
* two
* three

</v-clicks>

---

# Last slide
```
3. Launch locally and open the slides in your browser
4. You should see two errors:
  - On the first slide, the code blocks are highlighted in sequence (the `{at:0}` of the second block is ignored)
  - On the second slide, the `at="0"` of the `v-clicks` works fine: the lines appear when the code lines are highlighted. But you'll notice that 3 other clicks are required to go to the last slides. I'm wondering if it's related so I include this in the issue report.

 - OS: Ubuntu 22.04.3 (Under WSL with Windows 11)
 - Browser: Edge 120.0.2210.121
 - Slidev version: v0.46.1

Many thanks for your help (and for building an awesome product) !

KermanX:
## Overview

This PR refactors the clicks system. 

Affected:

- `v-click`, `v-after`, `v-click-hide` directives
- `VClicks`, `CodeBlockWrapper`, `KaTexBlockWrapper`, `SlidevVideo` components
- new `VClickGap` component

Fixes:

- fix #805
- fix #863
- fix #1136
- fix #1246
- fix #1231 (Likely, I can't fully understand this issue)

### Design

I personally think that the clicks system is a little ambiguous.

This PR considers the clicks to be whether **relative** or **absolute**.

```md
<div v-click>        relative   show >= 1 </div>
<div v-click="'+1'"> relative   show >= 2 </div>
<div v-click="4">    absolute   show >= 4 </div>
<div v-click-hide>   relative   hide >= 3 </div>

'''js {1|2}{finally:'all',at:'+1'【default】}
1                          highlight <  4 || > 4
2                          highlight == 4 || > 4
3                          highlight         > 4
'''

'''js {hide|2|3}{at:2【absolute】}
1                          hide < 3
2                          hide < 3, highlight == 3
3                          hide < 3, highlight >= 4
'''

<div v-after>        relative   show >= 4  </div>
<div v-click="10">   absolute   show >= 10 </div>
<div v-click="'+2'"> relative   show >= 6  </div>
```

#### `"relative" elements` vs. `previous auto-calculate clicks`

Similar to previous auto-calculate clicks, "relative" elements' click num are calculated depending on the total click num of "relative" elements before this element.

> Previously, all `v-click` elements, including those with an absolute click num, will add the auto-calculate clicks by one:
> ```markdown
> <div v-click="100"> I am absolute! </div>
> <div v-click> The auto-calculate click num is 2, instead of 1 </div>
> ```

In this PR, only "relative" elements get auto-calculate clicks and add the "relative"s' current total click num by its "size" (for `v-click` it's 1, for `v-click="'+3'"` it's 3, for `v-after` it's 0). The example is above.

### User-facing

- `VClicks.props.every` now accepts string.

- Code block/latex block line highlight range now supports `'none'` and `'hide'`.
  i.e. Accepts ranges like `1-3,5`, or `'none'`, or `'hide'`
  - `'none'`: No lines are highlighted
  - `'hide'`: The code block is hidden

- Code block/latex block line highlight adds a `finally` prop.
  Default: `'last'`, the last highlight range. If no highlight animation, all lines are highlighted.
  This can also be ranges like `1-3,5`, `'none'`, or `'hide'`.
  The difference between this prop and the last one in the ranges: See https://github.com/slidevjs/slidev/pull/1279#issuecomment-1933816902.

- The relative value of `at` prop like `"+1"` `"-1"` now affects the position of following "relative" elements. 
  See https://github.com/slidevjs/slidev/pull/1279#issuecomment-1934120532 and https://github.com/slidevjs/slidev/pull/1279#issuecomment-1935765920
  - `v-click`  ===  `v-click="'+1'"`
  - `v-after`  ===  `v-click="'+0'"`

- Add `v-after.fade` and `v-click-hide.fade` for consistency with `v-click`.

- <del> Deprecate `v-click.hide` (it works the same as `v-click-hide`, and there hasn't been a `v-after.hide`) </del>
  `v-click.hide` and `v-after.hide` are both available now.

- <del> Remove `clicks` in page frontmatter, because it is useless now </del> restored.

### Internal

- Removed `clicks`, `clicksDisabled`, `clicksElements` and `clicksOrderMap`
- Added `clicksContext` (see https://github.com/slidevjs/slidev/pull/1279#issuecomment-1936213651)

## Todos

- [ ] Documentation. Help wanted: Since I am not a native English speaker, it is difficult for me to describe the changes introduced by this PR in the most precise and concise terms.
- [x] add `v-after-hide` or fix `v-after` to support this?
```md
<div v-click-hide> Hide since 1 </div>
<div v-after>      Hide since 1 </div>
```
- [x] Fix `SlidevVideo` (which was introduced by #905, but I am wondering how it worked with clicks before)
- [x] Add `VClickGap` component for a gap in clicks flow
- [ ] (not necessary) <del> Find a custom directive hook that is first called in `parent`, then in `child` (https://github.com/vuejs/core/issues/10295) </del> This enables nested `v-click` directive, which is not supported previously.


antfu:
Hey, thanks a lot for the initiative! Indeed I think the clicks system needed some improvements.

Some feedback:

> `flow` value

I am not very sure what `flow` means exactly - but it seems to be pretty similar to the previous auto-calculate clicks. If it's the default, I wonder what's the reason/need to introduce this value?

> code block/latex block line highlight range now supports 'none' and 'hide'

I like it!

> code block/latex block line highlight adds a finally prop

Is this necessary? People could do that with `{1|2|*}` to have the `finally: all` behavior. I would avoid introducing too many concepts especially if the current one works well

> the relative value of at prop like "+1" "-1"is removed

I am not sure if it's internal. IIRC, it was a PR to add that support. If there is no strong technical reason, I would keep it to avoid making breaking changes

> deprecate v-click.hide (it works the same as v-click-hide, and there hasn't been a v-after.hide)

Why not add `.hide` to `v-after`. Again I'd like to avoid breaking existing slides

> remove clicks in page frontmatter, because it is useless now

This is definitely a no to me. I use it a lot in my slides. No matter how perfect the clicks inference we might have, we still need to have an escape hatch for users to manually specify the clicks they want. In addition, it's possible to have dynamic conditions like:

```html
<div :class='$clicks > 5 ? 'some-class' : 'another-class'" />
```

KermanX:
@antfu Thank you for your support and affirmation. 

> it seems to be pretty similar to the previous auto-calculate clicks

It is. The previous auto-calculate clicks also consider the elements with specified `"at"`.

About the meaning of `"flow"`, I am not very sure whether `"flow"` is the best name. I will try to explain it more clearly later.

> > code block/latex block line highlight adds a finally prop
> 
> Is this necessary? People could do that with `{1|2|*}` to have the `finally: all` behavior. I would avoid introducing too many concepts especially if the current one works well

Yes, it is. because `{1|2|*}` takes 2 clicks if it is in the "flow", but `{1|2}` only takes one. (p.s. the first one in `{1|2|*}` means "before clicks into the code block" both in the past and in this PR).

> I am not sure if it's internal. IIRC, it was a PR to add that support. If there is no strong technical reason, I would keep it to avoid making breaking changes

I considered it again and I now think this is meaningful. I will implement this maybe tomorrow (UTC+8).

> Why not add `.hide` to `v-after`. Again I'd like to avoid breaking existing slides.

Nice idea

> > remove clicks in page frontmatter, because it is useless now
> 
> This is definitely a no to me. I use it a lot in my slides. No matter how perfect the clicks inference we might have, we still need to have an escape hatch for users to manually specify the clicks they want. In addition, it's possible to have dynamic conditions like:

I forgot that it is possible to use click system programatically😢. I will revert the removement.

---

By the way, could you have a look at https://github.com/vuejs/core/issues/10295? This is needed in this PR to work with nested `v-click`s properly.

antfu:
> v-click="'+2'" works like transform: translateX(2) instead of margin-left: 2 in CSS.

I don't understand what do you mean here.

> `<div v-click="4">show >= 4 </div>`

Well, ofc. But absolute values are less flexible, especially if you have a lot of click elements in a single slide. Imagine you want to insert a new v-click before this, you have to manually bump 1 for every absolute clicks after that. In that case relative values could save a lot.

> ```html
> <div v-click>        show >= 1 </div>
> <div v-click="'+2'"> show >= 3 </div> !!!!!
> <div v-click>        show >= 4 </div> !!!!!
>  ```

I like that, and yes this is probably more intuitive. We could do that.

> v-click="'flow'" is the same as v-click="'+1'"

Sounds good, then we could avoid introduce the `flow` value. And essentially:

- `v-click` (default, +1)
- `v-click="+0"` == `v-after`

```html
<div v-click>        show >= 1 </div>
<div v-click="'+2'"> show >= 3 </div>
<div v-click="'-1'"> show >= 2 </div>
```

I like that consistency 👍

KermanX:
新年快乐~

The ba9d9ab9c160a4993c522772c592dfa0c1617732 commit refactored the clicks context, and fixed Presenter and Print mode.

packages/client/logic/clicks.ts
```diff
@@ -0,0 +1,109 @@
...
+/**
+ * 'flow' => '+1'
+ *   '+3' => '+3'
+ *   '-3' => '-3'
+ *    '3' => 3
+ *      3 => 3
+ */
+export function normalizeAtProp(at: string | number = 'flow') {
+  if (at === 'flow')
+    return '+1'
```
antfu, line 105: I tend to not introduce the special handling of literal `flow`, instead we could say the default is `+1`.

packages/client/logic/clicks.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (typeof at === 'string' && '+-'.includes(at[0]))
+    return at
+  return safeParseNumber(at)
```
antfu, line 108: We could throw a warning here if the input is not valid.

packages/client/builtin/KaTexBlockWrapper.vue
```diff
@@ -20,53 +20,76 @@ Learn more: https://sli.dev/guide/syntax.html#latex-line-highlighting
...
+  const at = props.at === 'flow' ? '+1' : props.at
+  const atNum = safeParseNumber(at)
+  const inFlow = typeof at === 'string' && '+-'.includes(at[0])
+  const flowSize = inFlow
+    ? atNum + props.ranges.length - 2
+    : 0
+  const start = inFlow
+    ? clicks.flowSum + atNum - 1
+    : atNum
+  const end = start + props.ranges.length - 1
```
antfu, line 66: This chunk of code as been duplicate in a few components and it mighjt be hard to sync the changes later one - maybe we could have `parseAtProp` retrns an object with extra info.

packages/client/internals/SlideWrapper.ts
```diff
@@ -1,26 +1,14 @@
...
-      default: () => [] as Element[],
-    },
-    clicksOrderMap: {
-      type: Map,
-      default: () => new Map<number, HTMLElement[]>(),
-    },
-    clicksDisabled: {
-      type: Boolean,
-      default: false,
+      type: Object as PropType<ClicksContext>,
```
antfu, line 10: Maybe let's name this prop `clicksContext` to be explicit and avoid confusing with the clicks `$clicks` which contains a number.

packages/client/internals/SlidesOverview.vue
```diff
@@ -138,7 +139,7 @@ watchEffect(() => {
               <SlideWrapper
                 :is="route.component"
                 v-if="route?.component"
-                :clicks-disabled="true"
+                :clicks="useFixedClicks(route, 99999)[1]"
```
antfu, line 142: I like this refactor!

packages/client/modules/directives.ts
```diff
@@ -10,10 +9,8 @@ import {
   CLASS_VCLICK_PRIOR,
   CLASS_VCLICK_TARGET,
   injectionClicks,
```
antfu, line 11: `injectionClicksContext`

packages/client/logic/clicks.ts
```diff
@@ -0,0 +1,109 @@
...
+      map.set(el, resolved)
+    },
+    unregister(el) {
+      flow.delete(el)
+      map.delete(el)
+    },
+    get flowSum() {
+      // eslint-disable-next-line no-unused-expressions
+      routeForceRefresh.value
+      return sum(...flow.values())
```
antfu, line 84: The `flow` map seems to be only used here. Maybe we could use the `[...map.values()].map(i=>i.flowSize)` directly and have a single source of truth?

packages/types/src/types.ts
```diff
@@ -74,3 +75,28 @@ export type PreparserExtensionFromHeadmatter = (headmatter: any, exts: SlidevPre
...
+
+export type ClicksFlow = Map<ClicksElement, number>
+
+export interface ResolvedClicksInfo {
+  max: number
+  flowSize: number
+  isCurrent?: ComputedRef<boolean>
+  isActive?: ComputedRef<boolean>
+  shows?: ComputedRef<boolean>
+}
```
antfu, line 89: I think to me `flow` is still an a bit ambgious word - maybe we could name the type as `absolute` / `relative` everywhere?

The interface could be changed to:

```suggestion
export interface ResolvedClicksInfo {
  type: 'absolute' | 'relative'
  max: number
  delta: number
  resolved: number | [number, number]
  isCurrent?: ComputedRef<boolean>
  isActive?: ComputedRef<boolean>
  isShown?: ComputedRef<boolean>
}
```

packages/client/logic/nav.ts
```diff
@@ -45,30 +60,10 @@ export const currentLayout = computed(() => currentRoute.value?.meta?.layout ||
...
-      clicks = 0
-    return clicks
-  },
-  set(v) {
-    queryClicks.value = v.toString()
-  },
-})
-
-export const clicksTotal = computed(() => +(currentRoute.value?.meta?.clicks ?? clicksElements.value.length))
+export const clicks = computed(() => usePrimaryClicks(currentRoute.value))
```
antfu, line 63: Let's name it `clicksContext` and bring back to number version of `clicks` - there are slides are still using `$slidev.nav.clicks` (it was the only way before we have `$clicks`)

packages/client/logic/nav.ts
```diff
@@ -45,30 +60,10 @@ export const currentLayout = computed(() => currentRoute.value?.meta?.layout ||
...
-    if (Number.isNaN(clicks))
-      clicks = 0
-    return clicks
-  },
-  set(v) {
-    queryClicks.value = v.toString()
-  },
-})
-
-export const clicksTotal = computed(() => +(currentRoute.value?.meta?.clicks ?? clicksElements.value.length))
```
antfu, line 68: We also need `clicksTotal`

packages/client/composables/useContext.ts
```diff
@@ -1,22 +1,16 @@
...
 export function useContext(
   route: ComputedRef<RouteLocationNormalizedLoaded>,
-  clicks: WritableComputedRef<number>,
 ): SlidevContext {
   const nav = useNav(route)
-  const navClicks = useNavClicks(clicks, nav.currentRoute, nav.currentPage)
   return {
-    nav: {
-      ...nav,
-      ...navClicks,
```
antfu, line 18: I feel we might still need this, that `clicks`, `clicksTotal`, `clicksContext` should resolved from injections (as we have primary clicks and static clicks now). Components calling this function should get the context to the state of that slide

antfu (commented):
Overall great refactors! Thanks a lot!

packages/types/src/types.ts
KermanX: I think it is a good idea to use "relative"&"absolute" instead of "flow".

Maybe it is better to change the interface to:

```ts
export interface ResolvedClicksInfo {
  max: number
  relativeDelta: number    // For absolute element, it is 0
  isCurrent?: ComputedRef<boolean>
  isActive?: ComputedRef<boolean>
  isShown?: ComputedRef<boolean>
}
```

Reason:

- Maybe there will be clicks elements that can't be resolved into one click or one range.
- The `delta` of absolute elements will not be used by the click system.
- A new `if-else` statement will be added if we use `ResolvedClicksInfo.type` to distinguish two types.

Is the new one OK?

packages/client/logic/clicks.ts
KermanX: I've just combined `flow` into `map`, but the e2e tests failed because of infinite recursive. This because (实在抱歉我没法用英文准确地表达下面这段话) `flow`是在运行时由v-click指令和组件的`onMounted`钩子一边添加新的元素，一边读取最新的累计数据的，**不能是响应式的，不然就会陷入死循环**；而`map`的数据是在所有点击元素都已经解析好之后才能用的，**应当是响应式的**。我暂时想不到把flow合并到map的办法😭

packages/types/src/types.ts
antfu: Sounds good to me!

packages/client/logic/clicks.ts
antfu: I will give it a try after you finish it, you could leave the two maps there for now

KermanX:
I've just found that HMR still breaks the clicks system. This is because the same `CodeBlockWrapper` is mounted two times with two different new elements every HMR. The `onMounted` hook is called twice, but only the second one's `onUnmounted` hook is called, which causes the problem. I am wondering if it is a Vue bug, but I can't create a minimal repro.

> Only the second one's element can be seen in the DOM tree after HMR.

> I've tried `const vm = getCurrentInstance()` + `onUnmouted(..., vm)`, but the issue still happens.

packages/client/composables/useNav.ts
```diff
@@ -32,6 +32,9 @@ export function useNav(route: ComputedRef<RouteRecordRaw | RouteLocationNormaliz
     route,
     path,
     total,
+    clicksContext,
+    clicks,
+    clicksTotal,
```
antfu, line 37: Is it ok to use global state rather than the one provided in injection? Wouldn't that make dynamic condition in overviews broken?

antfu:
That's a really nice PR + the description you have! Certainly don't worry about the English writing - you are doing great!

Overall looks good to me, let's merge it and ship as a beta to test out how it goes. Thank you!

KermanX:
Thank you for the encouragement. I will try my best to update the docs as soon as possible.

PR #1283, opened 2024-02-10
fix: `codemirror` import error (#1281)

### LINKED ISSUE #1281: Slides cannot be modified on the side integrated editor
**Describe the bug**
No matter whether an existing project or a new project, the integrated editor cannot be used. The slides' content or notes cannot be displayed in the browser. But the notes can be shown in the presentation mode. 

**To Reproduce**
1. Initialise a new project `npm init slidev@latest`
2. Open `localhost:3030` in the browser
3. click the editor button
4. See the error
image

 - OS: macOS 14.4
 - Browser: Firefox, Chrome and Safari
 - Slidev version: v0.47.4

KermanX:
fix #1281

This PR fixes #1281 by partly reverts 4ef7fc934586758d2c8bdaa8bb0afa6e4605fad2.

It is confusing that the `codemirror` package isn't explicitly excluded from `optimizeDeps`, but it is in fact not optimized (**only when `@slidev/cli` and `@slidev/client` are installed from npm registry. In this monorepo, it is OK**). And `codemirror@5` is a UMD package, but it is then wrongly imported as an ESM package, which causes the issue. This PR re-adds `codemirror` related packages to `optimizeDeps.include` array, and import `CodeMirror` as default import.

I am not sure whether other packages removed from the `include` list by 4ef7fc934586758d2c8bdaa8bb0afa6e4605fad2 (listed below) should be added again or not.

**Lines removed in 4ef7fc934586758d2c8bdaa8bb0afa6e4605fad2**:
```ts
import { dependencies } from '../../../client/package.json'
//      optimizeDeps: {
          include: [
            ...Object.keys(dependencies).filter(i => !EXCLUDE.includes(i)),
            'codemirror/mode/javascript/javascript',
            'codemirror/mode/css/css',
            'codemirror/mode/markdown/markdown',
            'codemirror/mode/xml/xml',
            'codemirror/mode/htmlmixed/htmlmixed',
            'codemirror/addon/display/placeholder',
            'prettier/plugins/babel',
            'prettier/plugins/html',
            'prettier/plugins/typescript',
            'mermaid/dist/mermaid.esm.min.mjs',
            'mermaid/dist/mermaid.esm.mjs',
            'vite-plugin-vue-server-ref/client',
          ],
```

**This PR re-adds these**:
```ts
const INCLUDE = [
  'codemirror',
  'codemirror/mode/javascript/javascript',
  'codemirror/mode/css/css',
  'codemirror/mode/markdown/markdown',
  'codemirror/mode/xml/xml',
  'codemirror/mode/htmlmixed/htmlmixed',
  'codemirror/addon/display/placeholder',
]
```

### DIFF
--- a/packages/client/setup/codemirror.ts
+++ b/packages/client/setup/codemirror.ts
@@ -1,6 +1,6 @@
 import type { Ref, WritableComputedRef } from 'vue'
 import { watch } from 'vue'
-import * as CodeMirror from 'codemirror'
+import CodeMirror from 'codemirror'
 import 'codemirror/mode/javascript/javascript'
 import 'codemirror/mode/css/css'
 import 'codemirror/mode/markdown/markdown'
--- a/packages/client/setup/codemirror.ts
+++ b/packages/client/setup/codemirror.ts
@@ -1,6 +1,6 @@
 import type { Ref, WritableComputedRef } from 'vue'
 import { watch } from 'vue'
-import CodeMirror from 'codemirror'
+import * as _CodeMirror from 'codemirror'
 import 'codemirror/mode/javascript/javascript'
 import 'codemirror/mode/css/css'
 import 'codemirror/mode/markdown/markdown'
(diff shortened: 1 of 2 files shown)


antfu:
So, the reason for removing the `optimizeDeps.include` is that it doesn't work without shamefully-hoist on pnpm.

Maybe we could use a runtime interop to work for both cases? Like

```ts
import * as _cm from 'codemirror'

const cm = interopDefault(_cm) as import('codemirror')

function interopDefault(mod: any) {
  return mod.default || mod
}
```

(FYI, I didn't test the code above)

KermanX:
This seems works:

```ts
import * as _CodeMirror from 'codemirror'

// eslint-disable-next-line ts/ban-ts-comment
// @ts-expect-error
const CodeMirror: typeof _CodeMirror = _CodeMirror.fromTextArea ? _CodeMirror : globalThis.CodeMirror
```

PR #1308, opened 2024-02-19
fix: package and path resolving

### LINKED ISSUE #1238: The request url is outside of Vite serving allow list. (x3)
**Describe the bug**
When using slidev as a global user binary (installed at `$HOME/.local/lib/node_modules/.bin/slidev` using `npm install --prefix $HOME/.local/lib/node_modules`), the `slidev` binary reports the following error:

```
The request url "/home/amin/.local/lib/node_modules/@slidev/client/main.ts" is outside of Vite serving allow list. (x3)
```

**To Reproduce**
1. Install `@slidev/cli`, `@slidev/theme-default` & `playwright-chromium` for the current unprivileged user at `$HOME/.local/lib/node_modules/`
2. Add the `$HOME/.local/lib/node_modules/.bin` in your `PATH` environment variable's directory list
3. Restart the shell
4. Try running `slidev`
5. See the error

 - OS:Archlinux
 - Browser: N/A
 - Slidev version: 0.46.1

**Rationale**

It would be great to be able to use Slidev as a global binary for quickly serving markdown slides instead of having to initialize a new project using `npm init slidev` or use `npx @slidev/cli` since it requires more than just `@slidev/cli` (for exporting slides for instance).

This would also allow one to be able to quickly add Slidev in their dotfiles in order to quickly reinstall an operating system and be ready presenting multiples folders without having to manually type `npm install` for each presentation folders.

### LINKED ISSUE #1278: Monaco is causing an infinite iframe recursion
**Describe the bug**
Monaco is causing an infinite iframe recursion

**To Reproduce**
I did a repository to isolate and reproduce the problem :
https://gitlab.com/pinage404/slidev_with_monaco/-/blob/main/slides.md

:warning: https://pinage404.gitlab.io/slidev_with_monaco :arrow_left: this tab is causing memory leak with an infinite iframe recursion :warning: 

1. Install slidev
2. Use a Monaco code block in your slide
3. In front matter, [enable Monaco even in the exported SPA](https://sli.dev/custom/config-monaco#exporting) :  `monaco: true`
4. Build to SPA : `slidev build`
5. Host the SPA with something that respect the `_redirect` file
6. Go to the SPA

Your memory will be eaten until crash or kill

 - OS: NixOS 23.11.20240124.a77ab16 (Tapir)
 - Browser: Firefox 121.0.1 (64 bits) & Chromium Version 120.0.6099.224 (Build officiel) (64 bits)
 - Slidev version: 0.46.3 & 0.47.4

Possible cause :
1. Export the slide to SPA `slidev build`
2. Host it
3. Go to the SPA
4. It will do a request to `/iframes/monaco/assets/index-z9jPtVTR.js` (or other hash) which is not in built the SPA

With a server that respect the `_redirect` file
The iframe's assets are not found, so redirected to the index page, which will load Monaco in an iframe where the assets are not found, so redirect to the index page ...

KermanX:
### Overview

fixes #1238
fixes #1278

Previously, the resolving utilities were messed up, and not all the cases (like global installation and `@/`) are handled properly. This PR fixes that by refactoring the path-resolving utilities.

### Changes

- Refactor the path resolving utilities
  - Remove unnecessary/duplicate code
  - Move all these utilities to `packages/slidev/node/resolver.ts`
- Fix the global installation path resolving
- Make `@/` work properly
- Make the resolving logic of themes and addons consistent, and more robust
- Noticeably improve the loading performance by avoiding unnecessary/duplicate path-resolving
- <del>Resolve `userRoot`, `cliRoot`, and `clientRoot` at module level. So that we don't need to pass them around anymore</del>
- Fix Monaco iframe builder

### Test Plan

- [x] Reload when theme changed
- [x] Reload when config changed
- [x] Create new slides.md by CLI
- [x] Theme HMR
- [x] `slidev`
- [x] `slidev build`
- [x] `slidev export` / `slidev export-notes`
  - Slidev installed globally (NPM&PNPM) & Playwright installed globally
  - Slidev installed globally (NPM&PNPM) & Playwright installed locally
  - Slidev installed locally (NPM&YARN&PNPM) & Playwright installed globally
  - Slidev installed locally (NPM&YARN&PNPM) & Playwright installed locally
- [x] `slidev format`
- [x] `slidev theme eject`
  - Theme installed as global package
  - Theme installed as local package
  - Theme is a local folder

### Todos

- [x] Get #1305 merged first


packages/slidev/node/resolver.ts
```diff
@@ -0,0 +1,222 @@
...
+  const npmPath = join(globalDirs.npm.packages, name)
+  if (fs.existsSync(`${npmPath}/package.json`))
+    return npmPath
+  if (ensure)
+    throw new Error(`Failed to resolve global package "${name}"`)
+}
+
+export const cliRoot = fileURLToPath(new URL('..', import.meta.url))
+export const clientRoot = await findPkgRoot('@slidev/client', cliRoot, true)
+export const userRoot = await getUserRoot()
```
antfu, line 71: I am not so sure about the top-level await here. Could we just run the getter when it's been used?

KermanX: `cliRoot` and `clientRoot` can be safely resolved at top-level.

About the `userRoot`, previously it was the `dirname(fullEntry)`, which didn't consider the `package.json`.

Currently `userRoot` is the "closest" package root, and `userWorkspaceRoot` is the "closest" workspace root (fallback to `userRoot` if there is no `pnpm-workspace.yaml`).

antfu: I am mostly concerned about performance and side-effects - usually a module should not contain top-level await which will block the whole module evaluation, especially this module exposes utilties - which means every module import those utils has to wait for the functions to be executed.

KermanX: I hadn't considered that before😭. I will fix this later.

PR #1322, opened 2024-02-21
feat: add `zoom` option in page frontmatter

### LINKED ISSUE #1022: Feature request: Hide scrollbar in iframe-* layout.

Hidden scrollbar from iframe (in iframe-* layout),  
or Markdown code section (seems useless?)

An option in iframe-* layout.

A hack for overflow the scrollbar element

```css
iframe {
    height: 102% !important;
}
```

### LINKED ISSUE #1075: (Enhancement) Variant on the <Transform> component
So, I try to get a lot of content on each of my slides, and so I've heavily relied on directives like `<Transform :scale=".9">` to get everything to fit.  But when I do so, I get vertical space (which is good) -- but the right margin moves inwards due to the scaling.

I've created a variation of `<Transform>` which adjusts the width inversely to the scale factor, which does exactly what I need.  So if scale==0.80, then width=="125%".  

But I'm thinking that most users of `<Transform>` actually want this variant? In any case, is this a reasonable widget to include in the main distribution?

p.s. "Pack" is a horrible name, but I'm usually using it to pack data onto a slide :-)

```vue
<!--
Apply scaling or transforming to elements.

Usage:

<Pack :scale="0.5">
  <YourElements />
</Pack>
-->

<script setup lang="ts">

const props = defineProps<{
  scale?: number | string
  origin?: string
}>()

const style = computed(() => {
  const transforms = []
  let width;
  if (props.scale != null) {
    transforms.push(`scale(${props.scale || 1})`)
    width = `${100.0/(props.scale || 1)}%`
  } else {
    width = "100%"
  }

  return {
    'width': width,
    'transform': transforms.join(' '),
    'transform-origin': props.origin || 'top left',
  }
})
</script>

<template>
  <div :style="style">
    <slot />
  </div>
</template>
```

KermanX:
This PR adds a `zoom` option in **every page**'s frontmatter to scale each slide **separately**.

| `zoom: 0.3` | `zoom: 1` (default) | `zoom: 3` |
| - | - | - |
| image | image | image |

Resolves #1075.

If we want to zoom all the slides, we can use the `canvasWidth` option in the headmatter, which has almost the same effect as applying this `zoom` option to all slides.

### DIFF
--- a/packages/client/constants.ts
+++ b/packages/client/constants.ts
@@ -38,6 +38,7 @@ const FRONTMATTER_FIELDS = [
   'src',
   'title',
   'transition',
+  'zoom',
 ]
 
 const HEADMATTER_FIELDS = [
--- a/packages/client/internals/SlideWrapper.ts
+++ b/packages/client/internals/SlideWrapper.ts
@@ -1,6 +1,7 @@
-import { defineComponent, h, provide, ref, toRef } from 'vue'
+import { computed, defineComponent, h, provide, ref, toRef } from 'vue'
 import type { PropType } from 'vue'
 import type { ClicksContext, RenderContext } from '@slidev/types'
+import type { RouteRecordRaw } from 'vue-router'
 import { injectionActive, injectionClicksContext, injectionCurrentPage, injectionRenderContext, injectionRoute } from '../constants'
 
 export default defineComponent({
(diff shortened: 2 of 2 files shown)


KermanX:
I've just tested it. It works well I think.

Issue #1207, opened 2023-12-04
Black background when exporting to PDF


piranna:
**Describe the bug**
I'm using `shibainu` theme, and when exporting to PDF both with the `download` frontmatter option or with `slidev export` command, generated PDF slides have a black background instead of the theme colorful ones. I think it can be related with the fact they are SVG images, so they should be rasterized to PNG images with the actual slides resolution first and use them instead. Rendering on web works properly.

**To Reproduce**
1. Create a new sli.dev with `shibainu` theme
2. Generate PDF with `slidev export`
3. Open the generated PDF
4. See the black background instead of the colorful one

**Desktop (please complete the following information):**
 - OS: Ubuntu Mate 23.10
 - Browser: Chrome 119
 - Slidev version: 0.44.0

piranna:
`bricks` theme has the same problem, polygons are shown, but background is not rendered since it's a SVG image, showing it a white background.

piranna:
Still happening in version 0.46.1.

AlbertBrand:
Try the following:

Open `node_modules/@slidev/cli/dist/export-WYWYD5MK.mjs` (this is for 0.46.3).
Find:
```
  const browser = await chromium.launch({
    executablePath
  });
```
(in the function `exportSlides`)

Change to:
```
  const browser = await chromium.launch({
    executablePath,
    headless: false,
  });
```

Then look for:
```
    await page.waitForLoadState("networkidle");
```

And add the following line after it:
```
    await page.waitForLoadState("networkidle");
    await page.pause();
```

Now run export. Playwright will run in headed mode and pause on the first page. You can use the Inspector panel to figure out which CSS is causing issues. 

I've seen an issue with page height, I needed to add:
```
.slidev-page {
  @apply h-full;
}
```
to my theme `layout.css` to make SVG size correctly.

longshihui:
same problem, Does this issue have a corresponding pull request?

PR #1340, opened 2024-02-27
fix: black background when exporting to PDF

kamuiiiii:
fix https://github.com/slidevjs/slidev/issues/1207

https://github.com/slidevjs/slidev/issues/1207 is caused by [_stacking context_](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_context#description)

### DIFF
--- a/packages/client/internals/PrintContainer.vue
+++ b/packages/client/internals/PrintContainer.vue
@@ -50,6 +50,6 @@ provideLocal(injectionSlideScale, scale)
 }
 
 .print-slide-container {
-  @apply relative overflow-hidden break-after-page;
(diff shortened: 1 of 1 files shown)


KermanX:
Thanks for your contribution!

I think the style should be like this (not 100% sure):

```vue
<style lang="postcss">
#print-content {
  @apply bg-main;
}

.print-slide-container {
  @apply relative overflow-hidden break-after-page translate-0;  // <--- Add `translate-0` to here
}
</style>
```

kamuiiiii:
Done, Appreciate your input

PR #1342, opened 2024-02-27
fix: monaco editor types resolution

KermanX:
This PR fixes:

- omit duplicate dependencies

- declaration files searching


packages/slidev/node/plugins/monacoTypes.ts
```diff
@@ -43,19 +43,20 @@ export function createMonacoTypesLoader({ userRoot }: ResolvedSlidevOptions): Pl
         ].join('\n')
       }
 
-      const matchLoad = id.match(/^\/\@slidev-monaco-types\/load(\/.*?)&name=(.*)$/)
+      const matchLoad = id.match(/^\/\@slidev-monaco-types\/load\/(.*?)&name=(.*)$/)
```
KermanX, line 46: Not very sure about this change.

antfu: Some resolution add leading slash some are not, you will see paths like:

`/@slidev-monaco-types/load//Users/xx...` and 
`/@slidev-monaco-types/load/Users/xx...`

We should probably unify them on generation. We just need to make sure the `root` passed to `fg` is absolute.

(and this might have some issues on Windows? 👀)

KermanX: Yes. This is broken on Windows.

KermanX: I've changed the URL format to 
- `/@slidev-monaco-types/resolve?pkg=...&importer=...`
- `/@slidev-monaco-types/load?root=...&name=...`

Also, raw imports are now like `...?monaco-types&raw`, so that `manualChunk` can work correctly.

PR #1353, opened 2024-02-28
ci: production smoke test

KermanX:
Not very sure about the name of this CI.

This PR adds a CI that simulates the user's behavior.

Conditions (`2x3x2 + 1x2x2 = 16`):
- NPM / PNPM / YARN
- Use a package and install Slidev locally / Use a single md file and install Slidev globally
- Hoist or non-hoist when installed locally
- Ubuntu or Windows

Note that MacOS is not included in this test, because it always OOM when building:
image

Also, YARN global install hasn't been included yet. It seems that YARN's global install is not very common.

Also, the following error happens randomly at a low frequency, but fixing this is out of this PR:
image

Visit https://github.com/KermanX/slidev-add-ci/actions to see the output of this CI in a copy of Slidev.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -16,12 +16,14 @@ A clear and concise description of what the bug is.
 
 **To Reproduce**
 Steps to reproduce the behavior:
+
 1. Go to '...'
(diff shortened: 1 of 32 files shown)


antfu:
Great! I think we could also try cypress e2e test against production build?

PR #1357, opened 2024-02-29
fix: make slide routes async, prevent crash on syntax errors

### LINKED ISSUE #1355: when I write '---' in slides.md file to add a page , this full page are reloaded, but I don't want to reload all pages, how to auto focus current page?
**Describe the bug**
when I write '---' in slides.md file to add a page , this full page are reloaded, but I don't want to reload all pages, how to auto focus current page?

**To Reproduce**
1. Go to write '---' on slides.md file , ctrl + s to save
2. the full page reload

 - OS: windows 11
 - Browser: chrome
 - Slidev version: latest

KermanX:
This PR contains:
- No white screen when there is a syntax error in slides, and auto-reload when the error is solved.
- Resolves #1355. Now adding and removing slides will not trigger a full reload.
- Refactors `useNav`, which used to be very ambiguous.
  - `useNav` used to mix the primary route and fixed route for exporting.
  - Now, it is renamed to `useNavBase` and `useFixedNav` (similar to the clicks context)
- Migrate to vite v5's HMR API, and give types to custom events.

### Internal changes
- Slide components are now async. And a loading animation is added after `100ms`. **Not very sure about this animation**.
- `rawRoutes` is renamed to `slides`, with reduced data.


packages/slidev/node/plugins/loaders.ts
```diff
@@ -154,12 +154,12 @@ export function createSlidesLoader(
         const moduleIds = new Set<string>()
 
         if (data.slides.length !== newData.slides.length) {
-          moduleIds.add('/@slidev/routes')
+          moduleIds.add('/@slidev/slides')
```
antfu, line 157: I searched a bit and seems there are quite many users are using it, I prefer to not rename it if there is no hard reasons

antfu (approved):
Very solid PR as always, thanks!

packages/slidev/node/plugins/loaders.ts
KermanX: Resoved via a1f67110a882457b9bf6d6b1fb10730f9b32f731.

Also, if a `routeAlias` is defined, the URL of the slide will always be the `routeAlias`. Previously, `routeAlias` is only a redirect.

PR #1365, opened 2024-03-03
refactor!: separating nav.ts for better scoping

antfu:
Move most of the logics out of `logic/nav.ts` with became a bit bloated overtime. This would help better maintainability in the long term

This can be a breaking change if you are relying on the exported props from `nav.ts` - basically renaming things, please refer to the types to migrate

### DIFF
--- a/packages/client/logic/nav.ts
+++ b/packages/client/logic/nav.ts
@@ -30,7 +30,7 @@ export const queryClicks = computed({
 export const {
   slides,
   total,
-  path,
+  currentPath,
   currentSlideNo,
   currentPage,
   currentLayout,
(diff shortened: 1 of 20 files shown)


packages/client/logic/nav.ts
```diff
@@ -64,206 +27,41 @@ export const queryClicks = computed({
...
+  prevSlide,
+} = useNavBase(
+  currentSlideRoute,
+  clicksContext,
+  queryClicks,
+  router,
+)
+
+watch([total, currentSlideNo], async () => {
+  if (hasPrimarySlide.value && !getSlide(currentSlideNo.value)) {
```
KermanX, line 63:
```suggestion
watch([total, route], async () => {
  if (hasPrimarySlide.value && !getSlide(route.value.params.no as string)) {
```

`currentSlideNo` is computed, and is always in the range of slides.

PR #1399, opened 2024-03-09
feat: Add guide to adding svg images

John98Zakaria:
Continues: https://github.com/slidevjs/docs/pull/172

Sorry I hadn't seen your response 3 weeks ago

> Slidev does not even have vite-svg-loader installed - how is this relevant?

I checked that you are right `pnpm why vite-svg-loader`, however the issue is still the same

I created a stackblitz demonstarting the issue

https://stackblitz.com/edit/slidev-azbcjy?file=slides.md,public%2Ftux.svg

If you try to `pnpm build` this it will fail.

If you apply the suggested fix it builds.

### DIFF
--- a/docs/guide/syntax.md
+++ b/docs/guide/syntax.md
@@ -365,6 +365,12 @@ For local assets, put them into the [`public` folder](/custom/directory-structur
 ![Local Image](/pic.png)
 ```
 
+In the case want to include an **SVG image** you need to add `?skipsvgo` to the image path.
+
+```md
+![Local Image](/pic.svg?skipsvgo)
+```
+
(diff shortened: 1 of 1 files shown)


antfu:
No I believe it's still not the case, we don't have svgo in the deps. The case is that the assets should start with `/` but nothing to do with `?svgo`

https://stackblitz.com/edit/slidev-eervwp?file=slides.md,package.json

(and the playground you send is using old version of Slidev)

John98Zakaria:
I wonder why, I wonder how I had this issue...

https://www.youtube.com/watch?v=wCQfkEkePx8

PR #1402, opened 2024-03-09
feat: draggable element

KermanX:
This PR implements a `<v-drag>` component:

```vue
<v-drag pos="451,398,239,48">
draggable box
<img ... >
</v-drag>
```

> Outdated. Please check the docs for the usage.

Then you can double-click the element in the slide, and then drag it to resize and move it. The `pos` prop will be auto-updated. Clicking outside the currently active fixed element will stop the dragging.

Initially, you don't need to specify the `pos` prop, which will be auto-generated on the first drag.

The format of `pos` is `left,top,width,height,rotate`. `rotate` is omitted when it is `0`.

Press <kbd>shift</kbd> when resizing, the aspect ratio will be fixed.

You can also use arrow keys to move the element.

image

### About HMR

In the following situations, the new position will be written back to the Markdown:

- Dragged && `debounce(500)`
- Click outside the element to stop dragging
- The window lost its focus

However, there are still some restrictions:

- When dragging pos isn't saved yet, modifying the Markdown file **via the external editor** will cause UB. In most cases, this would not happen because the window will lose its focus when you switch to the code editor.
- Multi-user editing at the same time will cause UB.

### Todos

- [x] Documentation. I will work on this after the functions get approved.
- [x] Find a way to support `<v-drag ...` in code block or comment. Currently, the writing back is based on the index of regex match.
- [x] Allow <del>`width` and</del> `height` to be `auto`.

### Notes

It would be better if we could use `v-fixed` as a directive, but I don't know how to do this.

The behavior is mostly learned from Figma.


packages/slidev/node/vite/loaders.ts
```diff
@@ -103,6 +103,8 @@ export function createSlidesLoader(
   let _layouts_cache_time = 0
   let _layouts_cache: Record<string, string> = {}
 
+  let skipHmr: [filePath: string, fileContent: string] | null = null
```
antfu, line 106: It can be a map or an array I guess?

packages/client/composables/useDragElements.ts
```diff
@@ -0,0 +1,71 @@
...
+    },
+    unregister(id) {
+      elements.splice(elements.indexOf(id), 1)
+    },
+    update(id, posStr) {
+      const idx = elements.indexOf(id)
+      if (idx < 0 || !info.value)
+        return
+      const oldContent = info.value.content
+      const match = [...oldContent.matchAll(/<(v-?drag)(.*?)(?:pos=".*?")(.*?)>/ig)][idx]
```
antfu, line 53: I guess it would be better to move these codemod logic to the server side.

antfu: Also, I think it might be better that we store the position in the frontmatter, that we could also assign IDs to v-drag and resolve the pos at runtime. For example:

```md
---
dragPos:
  foo: -8,134,153,48
---

<v-drag drag-id="foo">
foo
</v-drag>
```

This way we would be able to do positions even for `v-drag` inside components or some tricky places that regex couldn't cover.

KermanX: > Also, I think it might be better that we store the position in the frontmatter, that we could also assign IDs to v-drag and resolve the pos at runtime. For example:

I am wondering if this is the most convenient way. Users have to specify the ID by themselves, which can be quite annoying in my opinion. If we auto-generate IDs if no ID is provided, we have to use Regex again and the auto-generated IDs are not pretty. Since it is quite rare to use `v-drag` inside a component or in the code block, maybe we could give up this support as a trade-off.

antfu: I am thinking the most robust and generalized way, while you are considering the most user-friendly way - I think both are indeed valid perspectives.

The reason I lean more to robust is for long-term maintenance, the approach we are taken right now would lead to many false-positives we need to handle in the future. Also accidentally write positions to the wrong element might be annoying an confusing to users.

I guess we could support both - `pos` for in markdown notations, and `id` for v-drag in components, that we could throw errors when we failed to find the notation and ask users to explicit provide id when we failed to resolve the data, etc.

Also think we might be able to use a markdown-it plugin to track the source position more accurately.

KermanX: I've just renamed the types `DragElements***` to `DragElement***`, and allow users to neither specify `id` nor `pos` at the initial.

KermanX: btw, I am going to implement `height: unset` tomorrow, which is quite useful I think.

KermanX:
Is there any possibility we can use `VDrag` as a directive?

antfu:
Yeah I was thinking about that, having a directive would be nice but I guess it will not be easy to do. We could leave it for the future

antfu (approved):

Issue #1416, opened 2024-03-12
Exporting slides with v-clicks exports one slide per v-click even using --with-clicks false


oripka:
Is it possible that the new version and the v-click redesign, broke the export command? No matter what I pass as flags now it always exports a slide for every v-click

```
cd demo/starter && pnpm export --with-clicks false
```

e.g. slide 3 should not be there

To debug I went directly to: http://localhost:3030/print?print
Which also exhibits this behaviour.

I see some code in `export.ts` that simulates the clicks but I think that only applies if one uses `--per-slide`. This uses to work fine for me before updating to v0.48.

PR #1418, opened 2024-03-12
fix: Slide export without slide per click (#1416)

oripka:
Closes https://github.com/slidevjs/slidev/issues/1416 by disabling clicks when in print mode


KermanX:
I'm so sorry that I worked on this at the same time (#1419)

IIRC using `useNav()` inside `useClicks.ts` will cause HMR-related problems. So I think maybe #1419 may be a better fix.

Thanks for your contribution!

PR #1421, opened 2024-03-12
feat: PDF handout (slides on top, notes bottom of page)

oripka:
This pull request introduces a new option for slide export. It allows for the creation of an A4 page with the slide on the top and the speaker notes on the bottom.

The footer of the pages is customizable using a custom component `handout-bottom.vue`, similar to `global-bottom.vue` for slides.

Optionally, it can create cover pages by passing the `--cover` option, which then renders another custom component `handout-cover.vue`. Both of these components need to be in your theme or project root.

I have added some example components to the starter.

### Example Usage

```
cd demo/starter
node node_modules/@slidev/cli/bin/slidev.js export --handout --cover
```

The code creates 2 temporary files:

- A cover PDF (also accessible via /cover)
- A temporary handout PDF (also accessible via /handout), which forms the bottom part of the final handout

It then prepends the cover and merges slides and handouts. In this way, we can reuse the slide generation code. However, this breaks hyperlinks. Therefore, I have ensured to perform all necessary calculations to make them clickable also in the final PDF.

The final file will be: `handout-slides-export.pdf`

## Caveat

This feature requires the inclusion of this pull request to work properly: https://github.com/slidevjs/slidev/pull/1419
Without it, each click exports a slide, which then desynchronizes the merge with the notes causing errors.

### Terminology

Since the term "notes" was already taken, I called the API "handout", but it might be confusing, as /handout is just a helper page to generate the lower part of the final handout and will be merged with the slides to create the final handout.

### Fix `pageNumber` param

Somehow the pageNumber param does not get passed to HandoutBottom here: <HandoutBottom :pageNumber="index + 100" /> in PrintHandout.vue. This may be due to the way the component is imported. It would be nice to have, because for now, the page numbering is hardcoded in Slidev and not customizable.

### Configure output name

Allow configuring the output name of the handout

[handout-slides-export.pdf](https://github.com/slidevjs/slidev/files/14575271/handout-slides-export.pdf)

### DIFF
--- a/packages/client/composables/useNav.ts
+++ b/packages/client/composables/useNav.ts
@@ -65,6 +65,7 @@ export interface SlidevContextNavState {
   currentRoute: ComputedRef<RouteLocationNormalized>
   isPrintMode: ComputedRef<boolean>
   isPrintWithClicks: ComputedRef<boolean>
+  isHandout: ComputedRef<boolean>
   isEmbedded: ComputedRef<boolean>
   isPlaying: ComputedRef<boolean>
   isPresenter: ComputedRef<boolean>
@@ -237,7 +238,8 @@ const useNavState = createSharedComposable((): SlidevContextNavState => {
   const router = useRouter()
 
   const currentRoute = computed(() => router.currentRoute.value)
-  const isPrintMode = computed(() => currentRoute.value.query.print !== undefined)
+  const isPrintMode = computed(() => currentRoute.value.query.print !== undefined || isHandout.value)
+  const isHandout = computed(() => currentRoute.value.query.handout !== undefined || currentRoute.value.path.startsWith('/handout'))
   const isPrintWithClicks = computed(() => currentRoute.value.query.print === 'clicks')
   const isEmbedded = computed(() => currentRoute.value.query.embedded !== undefined)
   const isPlaying = computed(() => currentRoute.value.name === 'play')
--- a/packages/client/internals/PrintHandout.vue
+++ b/packages/client/internals/PrintHandout.vue
@@ -1,27 +1,25 @@
 <script setup lang="ts">
 import type { RouteRecordRaw } from 'vue-router'
 import { computed } from 'vue'
-
 import NoteDisplay from './NoteDisplay.vue'
 
 // @ts-expect-error virtual module
 import HandoutBottom from '/@slidev/global-components/handout-bottom'
 
 const props = defineProps<{ route: RouteRecordRaw; index: number }>()
-
 const route = computed(() => props.route)
 </script>
 
 <template>
   <div class="break-after-page">
-    <!--TODO: A4 specific, figure out better customization-->
+    <!--A4 specific, figure out better customization-->
     <div class="w-full mt-104 h-176 flex flex-col relative overflow-hidden">
       <NoteDisplay v-if="route.meta?.slide!.noteHTML" :note-html="route.meta?.slide!.noteHTML"
         class="w-full mx-auto px-2 handout-notes" />
 
       <div class="">
         <HandoutBottom  :pageNumber="index + 100" />
-        <!-- TODO: I would like to do this in HandoutBottom, but somehow props don't get passed. -->
+        <!-- I would like to do this in HandoutBottom, but somehow props don't get passed. -->
         <div class="absolute bottom-5 right-0 text-right text-[11px] ">
           {{ index + 1 }}
         </div>
--- a/packages/client/internals/PrintHandout.vue
+++ b/packages/client/internals/PrintHandout.vue
@@ -13,7 +13,6 @@ const route = computed(() => props.route)
 </script>
 
 <template>
-
   <div class="break-after-page">
     <!--TODO: A4 specific, figure out better customization-->
     <div class="w-full mt-104 h-176 flex flex-col relative overflow-hidden">
--- a/packages/client/pages/cover/print.vue
+++ b/packages/client/pages/cover/print.vue
@@ -1,0 +1,62 @@
+<script setup lang="ts">
+import { watchEffect } from 'vue'
+import { useNav } from '../../composables/useNav'
+import { themeVars } from '../../env'
+import { useStyleTag } from '@vueuse/core'
+
+const { isPrintMode } = useNav()
+
+// @ts-expect-error virtual module
+import HandoutCover from '/@slidev/global-components/handout-cover'
+
+watchEffect(() => {
+    if (isPrintMode)
+        (document.body.parentNode as HTMLElement).classList.add('print')
+    else
+        (document.body.parentNode as HTMLElement).classList.remove('print')
+})
+
+useStyleTag(`
+html.print,
+html.print body,
+html.print #app {
+  height: auto;
+  overflow: auto;
+}
+html.print #page-root {
+  height: auto;
+  overflow: hidden;
+}
+html.print * {
+  -webkit-print-color-adjust: exact;
+}
+html.print {
+  width: 100%;
+  height: 100%;
+  overflow: visible;
+}
+html.print body {
+  margin: 0 auto;
+  border: 0;
+  padding: 0;
+  float: none;
+  overflow: visible;
+}
+
+@page {
+  size: A4;
+  margin-top: 1.0cm;
+  margin-bottom: 1.0cm;
+  margin-left: 1.5cm;
+  margin-right: 1.5cm;
+}
+`)
+</script>
+
+<template>
+    <div id="page-root" class="grid grid-cols-[1fr_max-content]" :style="themeVars">
+        <HandoutCover />
+    </div>
+</template>
+
+<style lang="postcss"></style>
--- a/packages/slidev/node/cli.ts
+++ b/packages/slidev/node/cli.ts
@@ -535,96 +535,6 @@ cli.command(
   },
 )
 
-cli.command(
-  'export-handout [entry..]',
-  'Export handout to PDF',
-  args => exportOptionsHandout(commonOptions(args))
-    .strict()
-    .help(),
-  async (args) => {
-    const { entry, theme } = args
-    process.env.NODE_ENV = 'production'
-    const { exportHandout, getExportOptionsHandout } = await import('./commands/export')
-    const port = await getPort(12445)
-
-    for (const entryFile of entry as unknown as string) {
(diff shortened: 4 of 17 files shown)


packages/slidev/node/commands/export.ts
KermanX, line 1: Can we simplify the handout print logic here? There seems to be some duplicated code. And is that possible to have the cover and the content on a single page, so that we don't need to concat them and recalculate the links manually?

packages/client/composables/useNav.ts
```diff
@@ -241,7 +242,8 @@ const useNavState = createSharedComposable((): SlidevContextNavState => {
   const router = useRouter()
 
   const currentRoute = computed(() => router.currentRoute.value)
-  const isPrintMode = computed(() => currentRoute.value.query.print !== undefined)
+  const isHandout = computed(() => currentRoute.value.query.handout !== undefined || currentRoute.value.path.startsWith('/handout'))
+  const isPrintMode = computed(() => currentRoute.value.query.print !== undefined || isHandout.value)
```
KermanX, line 246: Could we use `/print?print=handout` instead of `/handout` as the path?

packages/slidev/node/virtual/global-components.ts
```diff
@@ -16,6 +16,18 @@ function createGlobalComponentTemplate(layer: 'top' | 'bottom'): VirtualModuleTe
               join(root, 'GlobalTop.vue'),
             ]
           }
+          else if (layer === 'handout-bottom') {
+            return [
+              join(root, 'handout-bottom.vue'),
+              join(root, 'HandoutBottom.vue'),
+            ]
+          }
+          else if (layer === 'handout-cover') {
```
KermanX, line 25: I am not sure whether `handout-cover` should be a global layer or a layout.

packages/client/pages/cover/print.vue
```diff
@@ -0,0 +1,62 @@
+<script setup lang="ts">
+import { watchEffect } from 'vue'
+import { useNav } from '../../composables/useNav'
+import { themeVars } from '../../env'
+import HandoutCover from '#slidev/global-components/handout-cover'
+
+const { isPrintMode } = useNav()
+
+watchEffect(() => {
+  const html = document.body.parentNode as HTMLElement
```
KermanX, line 10: I think maybe we don't need a `.print` class here. Because when this component is loaded, the page must be in the print mode.

KermanX (approved):
Thanks for your contribution! I think this feature is quite useful. btw, the commits I pushed earlier contain no actual function-related changes.

packages/slidev/node/commands/export.ts
oripka: I kept the cover and content separate because I use the same cover layout not only for a slide handout but also for a lab guide hand out (which is another PDF but generated from a Nuxt content site).

In general I like the idea of having the content on a single page. Now that you mention it I don't remember why I did it in two steps..

packages/client/composables/useNav.ts
oripka: I kept it separate in order to not mess with the existing components e.g. `PrintSlide.vue`, `PrintContainer.vue` but If we want to have it rendered in one step, it would make sense to have i in one route

oripka:
> Thanks for your contribution! I think this feature is quite useful. btw, the commits I pushed earlier contain no actual function-related changes.

Thanks for having a look at it and cleaning it up. I really appreciate it. I fetched your changes and seems to work. Thanks also for

packages/client/internals/PrintHandout.vue
```diff
@@ -13,13 +13,15 @@ const route = computed(() => props.route)
...
-        class="w-full mx-auto px-2 handout-notes" />
+      <NoteDisplay
+        v-if="route.meta?.slide!.noteHTML" :note-html="route.meta?.slide!.noteHTML"
+        class="w-full mx-auto px-2 handout-notes"
+      />
 
       <div class="">
-        <HandoutBottom :pageNumber="index + 100" />
+        <HandoutBottom :page-number="index + 100" />
         <!-- I would like to do this in HandoutBottom, but somehow props don't get passed. -->
```
oripka, line 25: I think we can remove the hard-coded page numbering then

packages/slidev/node/commands/export.ts
oripka: By the way I generate the lab guide from a nuxt content site and the issue is that the content is not really structured in pages (or slides). Alas a single nuxt content page can stretch over various print pages. So one does not know exactly where to do a page break. For this reason I just export all nuxt content articles to a single PDF (the browser automatically does the page breaks) and generate the footer with the page numbers separately using my slidev template. Of course this is very specific and certainly out of scope of Slidev. But I just wanted to mention my reason for implementing it in such a weird way ;)

KermanX:
Hello, there is an almost finished PR #1513 which allows users to provide their own print templates. Would that help?

oripka:
@KermanX I synced the current pull request with main. Unfortunately, I did not get around yet to implement the changes you mentioned.

oripka:
> Hello, there is an almost finished PR #1513 which allows users to provide their own print templates. Would that help?

Ok, thanks for the info, I will have a look how to implement it using this PR.

oripka:
> Hello, there is an almost finished PR #1513 which allows users to provide their own print templates. Would that help?

So I am starting to use this the mentioned PR, works but I have one question and one issue I am stuck with.

1. How can I debug export easier, before I could go to /handout in the browser (with my code) but now that the --template parameter was introduced I don't understand how to do live debugging, as --template is only accepted when exporting and not when running in dev mode (so /print in dev mode is always the default print), so one always has to do a new export run to see the result. Makes development a bit painful.

2. This is a template I am working on, I am stuck on how to properly scale the slide to fit to the A4 page with margins etc. it always ends up looking bad. I am sure I use the wrong way to do that (`slideWidth.value`, `provideLocal(injectionSlideScale, computed(() => 0.7))`.

Screenshot 2024-04-16 at 18 01 47



```
<script setup lang="ts">
import { useNav } from '../../composables/useNav'

import NoteDisplay from '../../internals/NoteDisplay.vue'
import { usePrintStyle } from '../../composables/usePrintStyle'
import PrintSlide from '../../internals/PrintSlide.vue'
import { provideLocal } from '@vueuse/core'
import { injectionSlideScale } from '../../constants'
import { useStyleTag } from '@vueuse/core'


const { slidesToPrint, total } = useNav()
import { slideWidth } from '../../env'

slideWidth.value = 740
provideLocal(injectionSlideScale, computed(() => 1))
</script>

<template>
    <div class="w-full px-[1px]"> <!--1px helps to avoid cutting off a tiny bit of content by margins-->
        <div v-for="(slide, index) of slidesToPrint" :key="slide.no"
            class="mx-auto w-full flex flex-col gap-4 break-inside-avoid-page break-after-page">
            <div>

                <div class="relative border-1 border-red w-full">
                    <PrintSlide :key="slide.no" :route="slide" class="mx-auto w-full border-0.5 border-black" />
                </div>

                <div class="flex flex-col w-full mt-12">

                    <div class="h-124 border-red border-2">
                        <NoteDisplay v-if="slide.meta.slide.noteHTML" :note-html="slide.meta.slide.noteHTML"
                            class="max-w-full h-full" />
                    </div>

                    <div class="w-full flex flex-col gap-2">
                        <div class="h-[1px] bg-black w-full" />
                        <div class="mt-2 relative">
                            <div class="top-3.0 absolute left-0 !text-[11px]">© {{ new Date().getFullYear() }} Company
                            </div>
                            <div class="top-3.0 absolute right-0 !text-[11px]">Title - Version</div>
                        </div>
                        <div class="text-right text-[11px] mt-6">
                        {{ index + 1 }} / {{ total }}
                    </div>

                    </div>
             
                </div>

            </div>
        </div>
    </div>
</template>

<style>
@page {
    size: A4;
    margin-top: 1.5cm;
    margin-bottom: 1cm;
}

/* overwrite the class in PrintSlideClick.vue, we want to break every slide
 * 794px fits A4
 * 1 px border (2x 0.5)
 * 26 px margin
 * 740 px slide width
*/
.print-slide-container {
    @apply relative overflow-hidden !break-after-avoid translate-0;
}


html.print,
html.print body,
html.print #app {
    height: auto;
    overflow: auto;
}

html.print #page-root {
    height: auto;
    overflow: hidden;
}

html.print * {
    -webkit-print-color-adjust: exact;
}

html.print {
    width: 100%;
    height: 100%;
    overflow: visible;
}

html.print body {
    margin: 0 auto;
    border: 0;
    padding: 0;
    float: none;
    overflow: visible;
}

@page {
    size: A4;
    margin-top: 1.5cm;
    margin-bottom: 1.0cm;
    margin-left: 1cm;
    margin-right: 1cm;
}
</style>

```

packages/client/setup/routes.ts
```diff
@@ -61,6 +61,16 @@ export default function setupRoutes() {
...
       },
+      {
+        name: 'handout',
+        path: '/handout',
+        component: () => import('../pages/handout/print.vue'),
+      },
+      {
+        name: 'cover',
+        path: '/cover',
+        component: () => import('../pages/cover/print.vue'),
```
antfu, line 72: Can this component be in `/handout` without being a route?

packages/slidev/node/cli.ts
```diff
@@ -611,6 +613,14 @@ function exportOptions<T>(args: Argv<T>) {
       type: 'number',
       describe: 'scale factor for image export',
     })
+    .option('cover', {
+      type: 'boolean',
+      describe: 'prepend cover to handout, needs handout-cover.vue in project',
+    })
```
antfu, line 619: I suppose we don't need this option? We could do it automatically when `handout-cover.vue` presents.

packages/slidev/node/cli.ts
```diff
@@ -611,6 +613,14 @@ function exportOptions<T>(args: Argv<T>) {
...
+    .option('handout', {
+      type: 'boolean',
+      describe: 'Export handout with slides on top and notes on bottom, optionally prepending a cover',
+    })
```
antfu, line 623: Could be `--type=handout` (default to slides). So later, we could merge `export-notes` to `--type=notes` and unify into a single command.

antfu: Oh I see @KermanX is doing `--template` in https://github.com/slidevjs/slidev/pull/1513

We could first discuss there

oripka:
I will close this pull requests in favor of a new one that uses the new export code

Issue #1428, opened 2024-03-14
Disable twoslash in dev mode


kravetsone:
**Describe the solution you'd like**
Add the option to disable twoslash for example using the cli or env argument for perfomance reasons

antfu:
We could use headmatters to do so as other options, like `twoslash: build` for example https://sli.dev/custom/#frontmatter-configures

PR welcome if you want to work on that

PR #1429, opened 2024-03-14
feat: option `twoslash` to disable twoslash

kamuiiiii:
Resolve https://github.com/slidevjs/slidev/issues/1428

### DIFF
--- a/docs/guide/syntax.md
+++ b/docs/guide/syntax.md
@@ -216,6 +216,14 @@ const count = ref(0)
 <!-- For the popup to not overlap the content below -->
 <div class="py-20" />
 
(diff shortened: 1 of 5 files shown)


KermanX (approved):

KermanX:
Maybe the docs should be updated too.

docs/guide/syntax.md
```diff
@@ -216,6 +216,14 @@ const count = ref(0)
 <!-- For the popup to not overlap the content below -->
 <div class="py-20" />
 
+#### Disable Twoslash
```
antfu, line 219: We don't need to do this here, mention it in the config.md would be better

PR #1435, opened 2024-03-15
feat: enhance `SlidevVideo` component

### LINKED ISSUE #1424: Export last frame of embedded videos
When including videos in my slides, only a black rectangle will be shown on the exported pdf.

For example:

Screenshot 2024-03-13 at 09 43 58

It would be amazing if instead, the last frame of the video was shown.

Untitled-1

It should be possible to manually extract the last frame from the video, or a video placeholder, and show this video place holder instead of the video when `isPrint` is `true`.

KermanX:
Resolves #1424.

> The following is the new doc for `SlidevVideo`

- `controls` (`boolean`, default: `false`): show the video controls
- `autoplay` (`boolean | 'once'`, default: `false`):
  - `true` or `'once'`: start the video only once and does not restart it once ended or paused
  - `false`: never automatically start the video (rely on `controls` instead)
- `autoreset` (`'slide' | 'click'`, default: `undefined`):
  - `'slide'`: go back to the start of the video when going back to the slide
  - `'click'`: go back to the start of the video when going back to the component's click turn
- `poster` (`string | undefined`, default: `undefined`):
  - The source of the image to print when the video is not playing.
- `printPoster` (`string | undefined`, default: `undefined`):
  - The override for `poster` when printing.
- `timestamp` (`string | number`, default: `0`):
  - The starting time of the video in seconds.
- `printTimestamp` (`string | number | 'last' | undefined`, default: `undefined`):
  - The override for `timestamp` when printing.

> [!WARNING]
> When exporting, the video may fail to load because Chromium does not support some video formats. In this case, you can specify the executable path of the browser. See [Chromium executable path](https://sli.dev/guide/exporting.html#executable-path) for more information.


antfu:
In a way, would having a video cover image be better? This option, to me, is a bit too specific and also might slow down the printing.

KermanX:
> In a way, would having a video cover image be better? This option, to me, is a bit too specific and also might slow down the printing.

Added an `imgToPrint` prop to achieve that. Also, I think most users may prefer waiting for a while when printing, rather than to screenshot by themselves...

QuentinRoy:
Thanks @KermanX  for this PR!
I have tried the new component on a project of mine (with default `"last"` timeToPrint) and it does not seem to be working. I still get the black rectangles. I am not sure yet if it is because it isn't supported by playwright on my laptop (macOS), or because of something else like the video not playing, or the renderer not waiting for it.

KermanX:
> I have tried your change in a project of mine with `timeToPrint` and it does not seem to be working 😥.

I had tested with a video that is not very long. I think the problem is the video not got fully loaded when exporting. I used to think that `waitFor('networkidle')` in the exporting should wait for the video, but maybe the video is mounted to DOM after this `waitFor` being called, which can be the reason.

KermanX:
The real problem seems to be that the video is just not being loaded in the print mode. The `@canplay` and `@metadataloaded` events both aren't fired, even after waiting for 30s. I am not sure why my previous video was printed correctly.

packages/client/builtin/SlidevVideo.vue
```diff
@@ -1,11 +1,14 @@
...
 import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
 import { useSlideContext } from '../context'
+import { useNav } from '../composables/useNav'
 
 const props = defineProps<{
   autoPlay?: boolean | 'once' | 'resume' | 'resumeOnce'
   autoPause?: 'slide' | 'click'
   autoReset?: 'slide' | 'click'
+  imgToPrint?: string
+  timeToPrint?: string | number | 'last'
```
antfu, line 11: I am wonder if we could have a general props like `poster`: https://www.w3schools.com/tags/att_video_poster.asp

I am not so sure about props to be print specific - I would probably use JS to do it conditionally.

KermanX:
> The real problem seems to be that the video is just not being loaded in the print mode. The `@canplay` and `@metadataloaded` events both aren't fired, even after waiting for 30s. I am not sure why my previous video was printed correctly.

It comes out that this is because chromium is missing some codecs. See https://github.com/microsoft/playwright/issues/4585.

If we run the export command with a Chrome binary, the video can be printed correctly:
```sh
pnpm export --executable-path "C:\Program Files (x86)\Microsoft\Edge\Application\msedge.exe"  
```

PR #1436, opened 2024-03-15
feat: enhance `RenderWhen` component

### LINKED ISSUE #1427: Detect whether exporting/presenting from a custom component
I want to implement custom progress bar component with different display logic when presentation is being presented or being exported. E.g. calculate progress position using timer value for presentation, but current/total slide number for exporting (see https://github.com/Envek/slidev-theme-envek/pull/1 for example)

Something similar to what [`RenderWhen` component](https://sli.dev/builtin/components#renderwhen) does ([source](https://github.com/slidevjs/slidev/blob/748f8bbc3589d2ac323d413103fec19bab13cda4/packages/client/builtin/RenderWhen.vue)) but for case when slide is being rendered for exporting.

Looked at `RenderWhen` component implementation, but can't see anything related to exporting in constants referenced by it. But maybe I'm just overlooking it…

KermanX:
resolve #1427.

- Add `print` context
- Add `#else` slot, which will be rendered if the context does not match.

### DIFF
--- a/packages/client/builtin/RenderWhen.vue
+++ b/packages/client/builtin/RenderWhen.vue
@@ -3,8 +3,9 @@ import type { RenderContext } from '@slidev/types'
 import { computed, ref } from 'vue'
 import { useElementVisibility } from '@vueuse/core'
 import { useSlideContext } from '../context'
+import { useNav } from '../composables/useNav'
 
-type Context = 'main' | 'visible' | RenderContext
+type Context = 'main' | 'visible' | 'print' | RenderContext
 
 const props = defineProps<{
   context: Context | Context[]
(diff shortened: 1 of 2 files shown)


packages/client/builtin/RenderWhen.vue
```diff
@@ -43,6 +47,8 @@ function conditionsMatch(context: Context) {
 <template>
   <div v-if="needsDomWrapper" ref="target">
     <slot v-if="shouldRender" />
+    <slot v-else name="else" />
```
antfu, line 50: Maybe `fallback`? That align with Vue better: https://vuejs.org/guide/built-ins/suspense.html#loading-state

PR #1448, opened 2024-03-20
chore: use `typescript@4` in `@slidev/client`

### LINKED ISSUE #1447: Unmet peer dependencies, @typescript/ata 0.9.4 x typescript@^4.4.4: found 5.4.2
**Describe the bug**

pnpm enables `strict-peer-dependencies` in the latest version, and `@typescript/ata 0.9.4` peers `typescript@^4.4.4`, but `slidev/client`'s typescript version is 5.4.2. so the start template installs deps via pnpm, will get a error.

```
 ERR_PNPM_PEER_DEP_ISSUES  Unmet peer dependencies

.
└─┬ @slidev/cli 0.48.3
  └─┬ @slidev/client 0.48.3
    └─┬ @typescript/ata 0.9.4
      └── ✕ unmet peer typescript@^4.4.4: found 5.4.2

hint: If you don't want pnpm to fail on peer dependency issues, add "strict-peer-dependencies=false" to a .npmrc file at the root of your project.
```

**To Reproduce**

1. pnpm create slidev@latest
2. use pnpm to install

**To Fix**

1. add `.npmrc` and set `strict-peer-dependencies=false` in the starter template.
2. or, fix the peer deps version

- OS: NA
- Browser: NA
- Slidev version: NA

KermanX:
fix #1447.

This PR pinned `typescript` to v4 in `@slidev/client` to meet `@typescript/ata`'s peer dependency version.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -57,7 +57,7 @@
     "recordrtc": "^5.6.2",
     "shiki": "^1.1.7",
     "shiki-magic-move": "^0.3.4",
-    "typescript": "^5.4.2",
+    "typescript": "^4.9.5",
     "unocss": "^0.58.5",
     "vue": "^3.4.21",
     "vue-demi": "^0.14.7",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -319,7 +319,7 @@ importers:
         version: 1.1.7
       '@shikijs/vitepress-twoslash':
         specifier: ^1.1.7
-        version: 1.1.7(typescript@5.4.2)
+        version: 1.1.7(typescript@4.9.5)
       '@slidev/parser':
         specifier: workspace:*
         version: link:../parser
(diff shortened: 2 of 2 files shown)


antfu:
I don't think this a good idea, the downgrade is never a long-term plan, and this would create misalignment with users' local TS, and we will never get future updates.

https://github.com/slidevjs/slidev/issues/1447 is honestly not a big deal to me, it's more like ATA should bump the deps rather than issues on us. At last, we could fork it - but honestly, I don't think it's Slidev's responsibility either.

KermanX:
I think the problem is that, as reported in #1447, the latest pnpm blocks users from installing. So we have to either downgrade our version, or fork ata or set `strict-peer-dependencies=false` in the template.😭

antfu:
Isn't that supposed to be a warning? In that case I'd prefer to add `strict-peer-dependencies=false` for now.

KermanX:
> Isn't that supposed to be a warning? In that case I'd prefer to add `strict-peer-dependencies=false` for now.

I haven't tested by myself, but according to the output of PNPM in #1447, it is an error (`fail`)

KermanX:
I found out why. `strict-peer-dependencies` is enabled by default since v7 (https://github.com/pnpm/pnpm/pull/4427), but re-disabled in `v7.3.15` (https://github.com/pnpm/pnpm/pull/5499).

PR #1453, opened 2024-03-20
Remove redundant onMount

splincool:
I have found redundant piece of code with empty `onMount() {}`


KermanX:
Thanks for your contribution! However, `onMount` is assigned in https://github.com/slidevjs/slidev/blob/main/packages/client/composables/useNav.ts#L298 .

Issue #1452, opened 2024-03-20
Add `startClicksCount` to frontmatter options


kravetsone:
**Is your feature request related to a problem? Please describe.**
This is convenient for reused slides with clickable components

Useful for combination with [multiple-entries](https://sli.dev/guide/syntax#multiple-entries) and
[custom-total-clicks-count](https://sli.dev/guide/animations#custom-total-clicks-count)

Usage example:

```yaml
---
src: ./slide-with-clicks.md
clicks: 4
---

---
src: ./another-slide.md
---

---
src: ./slide-with-clicks.md
clicks: 10
startClicksCount: 5
---

---
src: ./another-slide.md
---
```

**Describe the solution you'd like**
Add `startClicksCount` to frontmatter options

**Describe alternatives you've considered**
Idk

antfu:
That's an interesting usage - I don't mind to have it if it's not technically hard. Maybe name it `clicksStart` that could be visually grouped together with `clicks`?

kravetsone:
> That's an interesting usage - I don't mind to have it if it's not technically hard. Maybe name it `clicksStart` that could be visually grouped together with `clicks`?

It's a good name! Does anyone want to do this? I'm just not sure I'm that good at figuring out how to do it) Especially since I don't have much time

I am preparing my first talk and slidev helps me a lot! I want to continue using it)

You are amazing!

kravetsone:
> That's an interesting usage - I don't mind to have it if it's not technically hard. Maybe name it `clicksStart` that could be visually grouped together with `clicks`?

It's a good name! Does anyone want to do this? I'm just not sure I'm that good at figuring out how to do it) Especially since I don't have much time

I am preparing my first talk and slidev helps me a lot! I want to continue using it)

You are amazing!

kravetsone:
> That's an interesting usage - I don't mind to have it if it's not technically hard. Maybe name it `clicksStart` that could be visually grouped together with `clicks`?

it is posible to implement this without frontmatter options?
like workaround

PR #1456, opened 2024-03-21
feat: add `clicksStart` option in page frontmatter

kamuiiiii:
resolve https://github.com/slidevjs/slidev/issues/1452

### DIFF
--- a/packages/client/composables/useClicks.ts
+++ b/packages/client/composables/useClicks.ts
@@ -60,7 +60,7 @@ export function createClicksContextBase(
       // eslint-disable-next-line no-unused-expressions
       routeForceRefresh.value
       return clicksTotalOverrides ?? Math.max(0, ...[...map.values()].map(v => v.max || 0))
-    }
+    },
   }
 }
 
@@ -72,6 +72,6 @@ export function createFixedClicks(
   return createClicksContextBase(
     ref(Math.max(currentInit, clicksStart)),
     clicksStart,
-    route?.meta?.clicks
+    route?.meta?.clicks,
   )
 }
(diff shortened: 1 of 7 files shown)


KermanX:
Thanks for your contribution! Could we only change the following two places if possible?

https://github.com/slidevjs/slidev/blob/4a8ad4e56b6a08da64aa2a8fa0207c4c178b0e71/packages/client/composables/useNav.ts#L281-L288

https://github.com/slidevjs/slidev/blob/4a8ad4e56b6a08da64aa2a8fa0207c4c178b0e71/packages/client/composables/useNav.ts#L297-L301

kamuiiiii:
I've simplified the code. To ensure that 'clicks' and 'clicksStart' work together, I also believe that modifying 'nextSlide' and 'prevSlide' might be necessary.

KermanX (approved):

kamuiiiii:
Submitted updates to fix cases where clicksStart is greater than total.

packages/client/composables/useClicks.ts
```diff
@@ -16,10 +16,10 @@ export function createClicksContextBase(
   return {
     get current() {
       // Here we haven't know clicksTotal yet.
-      return Math.max(+current.value, clicksStart)
```
KermanX: I am not very sure if this is OK. Because `current` can be accessed at any time, but `this.total` is only available after the whole slide has been mounted.

kamuiiiii: I think accessing it in `current` is just like anywhere else the existing `total` only guarantees the correct behavior of the final application.

antfu (approved):

PR #1475, opened 2024-03-28
feat: context menu

KermanX:
This PR implements a context menu, which can be used in the play mode.

### Features

- Users can customize the menu items via `setup/context-menu.ts`
- There are three kinds of menu items:
  - Large (Icon + label)
  - Small (Icon only, label will show up when hovering)
  - Separator
- Press `shift` while right-clicking, and the context menu provided by the browser will show, instead of this one.
- `contextmenu: boolean | 'dev' | 'build' ??= true` in headmatter to enable / disable this feature.
 
### Why?

- Some devices cannot hover the left bottom part, but can only right-clicking. e.g. Blackboard computers.
- More familiar UX for users migrated from Microsoft PowerPoint.

### Todos

- [ ] Better default options
- [x] Docs

> [!WARNING]
>
> The major disadvantage of this feature is that this prevents the default context menu, although we can still press `shift` while right clicking to open it, it becomes less convenient.


antfu:
I think it would make sense to enable the custom context menu by default in 0.49, with an additional item "Always Use Native Context Menu" at the end that writes back to the headmasters to disable it. This way for ppl who prefer the previous behavior, they can get it back easily without reading the docs.

packages/slidev/node/utils.ts
```diff
@@ -23,30 +23,40 @@ export function generateGoogleFontsUrl(options: ResolvedFontOptions) {
   return `https://fonts.googleapis.com/css2?${fonts}&display=swap`
 }
 
-export function updateDragPos(slide: SlideInfo, dragPos: Record<string, string>) {
+/**
+ * Update frontmatter patch and preserve the comments
+ */
+export function updateFrontmatterPatch(slide: SlideInfo, frontmatter: Record<string, any>) {
```
antfu, line 29: Made this generalized for any frontmatter fields

packages/client/internals/ContextMenu.vue
```diff
@@ -0,0 +1,110 @@
...
+  return y
+})
+</script>
+
+<template>
+  <div
+    v-if="currentContextMenu"
+    ref="container"
+    :style="`left:${left}px;top:${top}px`"
+    class="fixed z-100 w-60 flex flex-wrap justify-items-start p-1 animate-fade-in animate-duration-100 backdrop-blur bg-main bg-opacity-75! border border-main rounded-md shadow overflow-hidden select-none"
```
antfu, line 64: Updated some styles in this component. Basically, we have pre-defined tokens like `border-main`, `bg-main`, `bg-active` etc.

packages/client/internals/ContextMenu.vue
```diff
@@ -0,0 +1,110 @@
...
+        <div class="mx-auto">
+          <component :is="item.icon" />
+        </div>
+        <div v-if="typeof item.label === 'string'">
+          {{ item.label }}
+        </div>
+        <component :is="item.label" v-else />
+      </div>
+    </template>
+    <template v-if="!isExplicitEnabled">
```
antfu, line 94: The `contextMenu` now defaults to `undefined` - set `true` explicit will hide this section

antfu (approved):

PR #1481, opened 2024-03-30
fix: v-motion

KermanX:
### This PR includes:

- No longer need `preload: false` when using `v-motion`
- Allow to combine click animation with `v-motion` via the `at` property in the variant.

### Example

```
<div
  v-motion
  :initial="{ x: 100, y: 0 }"
  :enter="{ x: 400, y:0}"
  :state-1="{ at: true, x: 400, y:100}"
  :state-2="{ at: '+1', x: 400, y:200}"
  :leave="{ x: 800, y:400 }"
  w-30 b b-white>
  clicks {{ $clicks }}
</div>
```

The meanings of variants:

- `initial`: `currentPage < thisPage`
- `enter`: When `currentPage === thisPage`,  it is a fallback if none of the `state-x` satisfies
- `state-x`: `x` should be a number. We can use `at` property here to specify the click
- `leave`: `currentPage > thisPage`

Note that `state-x` starts matching from the largest `x`. `x` can be considered as its priority


demo/starter/slides.md
```diff
@@ -1,394 +1,29 @@
 ---
-# try also 'default' to start simple
```
antfu, line 2: Looks good, tho we might want to revert the slide change

KermanX:
I've just found some designs that may be better. (Too busy these days😭)

PR #1492, opened 2024-04-02
fix: global layers

### LINKED ISSUE #1490: Slide export doesn't handle `v-if` in `global-top` or `global-bottom` based on conditional using `currentPage`, `currentLayout`
**Describe the bug**

The exporter seems to fail to handle `v-if` conditionals in `global-top` or `global-bottom` that use `currentPage`, `currentLayout`, and other navigation composables.

My global bottom is intended to only be displayed on non-`cover` slides after the first. This functions properly in browser view but fails to display properly on export.

```vue
<script setup lang="ts">
// import { computed } from "vue";
import { useNav, useSlideContext } from "@slidev/client"

// const { $slidev, $route } = useSlideContext()
const { currentPage, currentLayout } = useNav()
</script>

<template>
  <div v-if="currentPage.value > 1 && currentLayout !== 'cover'" class="slide-footer absolute w-full bottom-0 left-0 p-0 text-sm" style="display: grid; grid-template-columns: 33.33% 33.33% 33.34%; color: #ffffff">
    <div class="footer-comp justify-start" style="display: flex; flex-direction: row; gap: 7px; padding-left: 10px">
      <Item text="Ani Perumalla">
        <carbon:user-avatar-filled-alt />
      </Item>
      <Item text="ani-per.gitlab.io">
        <carbon:link />
      </Item>
    </div>
    <div class="footer-comp justify-center" style="display: flex; flex-direction: row; gap: 4px; padding-top: 4px; padding-bottom: 4px;">
      <img src="./img/logos/PSU_Engineering.png" alt="PSU Mark">
      <img src="./img/logos/PSU_Aero.png" alt="PSU Mark">
    </div>
    <div class="footer-comp justify-end" style="display: flex; flex-direction: row; gap: 7px; padding-right: 10px">
      <Item text="PSU UAS Research Lab">
        <carbon:drone />
      </Item>
      <Item text="<a href='https://purl.psu.edu'>purl.psu.edu</a>">
        <carbon:link />
      </Item>
    </div>
  </div>
</template>
```

This produces a proper global bottom in browser view, but no global bottom in the exported pdf:

Correct browser view

Incorrect exported pdf slide

When I remove the conditional `v-if="currentPage.value > 1 && currentLayout !== 'cover'"` from the first `div` in the template, the exporter displays the global bottom correctly:

Correct exported pdf slide

**Desktop**

- OS: Windows 11
- Browser: Brave, Firefox, Edge
- Slidev version: 0.48.8

KermanX:
This PR:

- fixes #1490
- Simplify code
- Now global layers are also displayed in the quick overview and preview next


antfu:
It breaks my usage of having a single instance of global layers. One of the point of global layers was to live out of slides lifecycle. I think maybe we should still keep the original behavior for playback but only duplicates in print mode (I use the `--per-slide` flag to workaround it in export, btw).

KermanX:
As mentioned in https://sli.dev/custom/global-layers#example, the global layers are used as something like footers. For footers they don't need to be singleton, but should be displayed with each slide. Should we add `SlideTop` and `SlideBottom` for footers, while preserving the current usage of global layers? If there are the `SlideTop` and `SlideBottom` components, I think the global layers are intended to be used as something like a navigator or so, which should not be printed.

PR #1493, opened 2024-04-02
fix: revert `overflow-hidden` for `SlideWrapper`

murilo-cunha:
Hi, thanks a lot for this project! Slidev's my go-to tooling these days. Big fan of the project!

This PR:
- Allow overflow to be visible to enable negative margins for styling

I like to style my slides with negative margins (especially when using `two-cols`, etc.) but with the change introduced in https://github.com/slidevjs/slidev/commit/c820e23b10581191a42bbf51c04bfd2caa49077a, it was hard to customize it to my liking.

Not sure there are any other side effects to this change, but it took me some time to fix it so I figured I'd open a PR :)

My current workaround was to include in `styles/main.css`:

```css
.slidev-page {
  /* https://github.com/slidevjs/slidev/commit/c820e23b10581191a42bbf51c04bfd2caa49077a */
  overflow: visible !important;
}
```

### DIFF
--- a/packages/client/internals/SlidesShow.vue
+++ b/packages/client/internals/SlidesShow.vue
@@ -76,7 +76,6 @@ function onAfterLeave() {
         :class="getSlideClass(route)"
         :route="route"
         :render-context="renderContext"
-        class="overflow-hidden"
       />
     </div>
   </component>


antfu:
Can you elaborate more what you want to do and why is this needed?

murilo-cunha:
In general so I can use negative margins to adjust the content.

Sometimes I like to have images partially over/covering each other in my slides (when using two-cols layout.

Or to make quick adjustments on the spacing for text.

I'm not sure if the overflow-hidden has other side effects, but I was using it as a quick way to increase the spacing of elements via negative margins

Issue #1010, opened 2023-05-29
Docker run not working


andradei:
Following the documentation here: https://sli.dev/guide/install.html#install-on-docker

I ran the command below (added `--platform`)

```
docker run --name slidev --rm -it \
          --user node \
          --platform linux/amd64/v3 \
          -v (pwd):/slidev \
          -p 3030:3030 \
          tangramor/slidev:latest


Unable to find image 'tangramor/slidev:latest' locally
latest: Pulling from tangramor/slidev
c41833b44d91: Pull complete
762c2470eea4: Pull complete
fefc7d195eee: Pull complete
06fc22ed341f: Pull complete
768063141391: Pull complete
3837a9825463: Pull complete
78cf9c1f14c9: Pull complete
Digest: sha256:9cd1c610d197b0a343c0bea9a3049af709f779496e67242e5ec8e9eee51aac09
Status: Downloaded newer image for tangramor/slidev:latest


docker: image with reference tangramor/slidev was found but does not match the specified platform: wanted linux/amd64/v3, actual: linux/arm64/v8.
See 'docker run --help'.
```
**Desktop (please complete the following information):**
 - OS: Linux
 - Browser: Brave
 - Slidev version: Docker (latest)

andradei:
It's fine that it's no priority to address this issue or keep the docker in prime shape. Hopefully there isn't a rule/config that auto-closes issues that haven't been addressed.

mikelorant:
You can't just add a platform, it needs to be built for that platform. However, have a look at the tags that exist:

https://hub.docker.com/r/tangramor/slidev/tags

You should be able to do a `docker pull tangramor/slidev:arm64v8` to get an `arm64` image.

An official image needs to be built containing multiple architectures and tagged with the release version. Relying on `latest` is very poor practises. As much as the documentation states this image is official, there is absolutely nothing about it that meets those requirements.

It would be easier to clone this repository and create your own image if you need an current `arm64` image.

This issue is user error, and would recommend closing it.

andradei:
@mikelorant I'm not trying to build it for arm. The `linux/amd64` image already exists and that's what should work by following the documentation. The error message is a strange one seen I'm not running an ARM CPU either.
I might still be very possible that I made a mistake somewhere, but I don't have issues with any other docker image I try to pull and build on the host machine.

PR #1511, opened 2024-04-10
docs: fix typo in exporting.md

FlyingCaiChong:


KermanX:
According to #1010, the Docker image seems to not work now. Should we change the docs?

PR #1513, opened 2024-04-10
feat: print templates, slide layers and rewrite export function

### LINKED ISSUE #1509: RFC: Custom UI layouts
One of Slidev's main features is to be hackable. It would be useful if UIs, especially the presenter's view (#1095), could be configured by users.

Also, `global-top` and `global-bottom` can be one of these UIs. (Related to #1492)

We may need a new folder like `pages` to store these UIs. For example, users can create a `pages/presenter.vue` of a component receiving the `#thisSlide`, `#previewNext`, `#note`, and `#editor` slots.

> edit: I am working on this now

KermanX:
close #1509

This PR:
- Allow user to customize print template via the `--template` or `-t` option. In dev mode, the option is called `--print-template`.
  - User can add or override print template via `./pages/print/*.{vue,ts}`
  - Print page is no longer being bundled in the build mode
  - The `export-notes` sub-command is removed. Use `-t notes` instead.
  - Previously, `-t` was the alias of `--theme`, but `--theme` is quite uncommon.
- Add the "handout" print template (close #1421)
- Add new layers for **each** slide:
  - `layouts/slide-top.vue`
  - `layouts/slide-bottom.vue`
- Refactors DOM structure

> The locations of the files and option names need some discussion.


docs/guide/install.md
```diff
@@ -202,7 +203,8 @@ Options:
 - `--range` (`string`): page ranges to export (example: `'1,4-5,6'`).
 - `--dark` (`boolean`, default: `false`): export as dark theme.
 - `--with-clicks`, `-c` (`boolean`, default: `false`): export pages for every click animation (see https://sli.dev/guide/animations.html#click-animations).
-- `--theme`, `-t` (`string`): override theme.
+- `--theme` (`string`): override theme.
+- `--template`, `-t` (`string`, default: `default`): specify [the print template](/guide/exporting#print-template).
```
antfu, line 207: I would prefer `--type` with hard-coded options like `slides` `notes` `handout` - and let ppl to override them instead of creating arbitrary options. Because we might want to do some special handling for each type - over-generalize it might not be ideal for long term.

KermanX: I think it is OK to allow users to create new print templates. This is because it will be the user/addon's responsibility to handle other options like `--with-clicks` or so.

About the name of this option, should we change all "print templates" into "print types" or only the CLI flag?

KermanX:
> In order to make the handout #1421 pull request work with this code I found that I need to be able to control the scaling factor (zoom) inside [SlideWrapper.vue](https://github.com/slidevjs/slidev/blob/6428d19273e1ec7b7ae60f0c4828ae84d5c487f8/packages/client/internals/SlideWrapper.vue)
> 
> Currently it is retrieved from the front matter
> 
> ```
> const zoom = computed(() => props.route.meta?.slide?.frontmatter.zoom ?? 1)
> ```
> 
> If that would be configurable from a template (e.g. via inject) it would solve the problem that I ran into when implementing a notes template ([point 2](https://github.com/slidevjs/slidev/pull/1421#issuecomment-2059443047)) where one adjusts the width and height correctly to fit an A4 page but the scale of the content is not adjusted and the content overflows out of the slide.
> 
> Here is a current draft of the [notes template](https://gist.github.com/oripka/fda08bfce37b6653c5e0ce1dd7d0a5be)

About the debugging issue you said in #1421, there is a `--print-template` option in this PR now.

About the slide scale, we can use `<SlideContainer>` here (I will do this change).

oripka:
I tried the latest commit. Seems like solving a lot of issues.

One thing I noted is that the page aligns well in Firefox but not in Chrome and when exporting on the command line.

Here is firefox on the left, chrome on the right

image

And a document `npx slidev export -t handout`

[slides-export-handout.pdf](https://github.com/slidevjs/slidev/files/15028601/slides-export-handout.pdf)

Maybe a bug in Chrome or a DPI issue?

oripka:
Doing a bit more testing having a slide like this with transitions and renderwhen also results in an empty slide.

```

---


# Test

<RenderWhen context="visible">
    <transition appear name="outro">
<img class="h-80 mx-auto" src="https://github.com/slidevjs/themes/blob/main/screenshots/theme-seriph/01.png?raw=true" alt="">
    </transition>
</RenderWhen>
<!--Test-->

<style>
.outro-enter-from {
  opacity: 0;
  transform: translate(-40px, 40px);
}
.outro-enter-active {
  transition: all 0.5s ease-in-out;
}
.outro-enter-to {
  transform: translate(0, 0);
}
.slidev-layout.outro {
  @apply bg-bottom bg-right;
}

</style>
```

image

KermanX:
I spent 2 hours on this, trying to figure out why the page is (exactly) 1.5x larger than the expected size, but made no progress. The output is still like this:

image

oripka:
> I spent 2 hours on this, trying to figure out why the page is (exactly) 1.5x larger than the expected size, but made no progress. The output is still like this:

Yes I have also spend some time debugging. I have found that changing the scale factor to 1.5 here works with your handout code, when exporting:
```
  async function genPagePdfOnePiece() {
    await go('print')
    await page.pdf({
      path: output,
      scale: 1.5,
      margin: {
        left: 0,
        top: 0,
        right: 0,
        bottom: 0,
      },
      printBackground: true,
      preferCSSPageSize: true,
    })
```

It might have to do with the DPI setting when exporting PDFs, here is a gist for debugging that tries to calculate the width and height according to the DPI (which is hardcoded for now): https://gist.github.com/oripka/36f4589df0fa72ed5758261c0a870264

KermanX:
> t might have to do with the DPI setting when exporting PDFs

I tried to set this setting to `100%` and restart the browser:

image

, but the page is still 1.5x larger. I am worried that if we hard-code `1.5` in the code, it may not work on some machine🥺

oripka:
Maybe this is related to these known issues (pupeteer and chrome):

https://issues.chromium.org/issues/40144973#comment9
https://github.com/puppeteer/puppeteer/issues/666
https://github.com/puppeteer/puppeteer/issues/2278

I tried this code to determine it dynamically, but I think this is the incorrect DPI for printing, as this is one just displaying the page. In my case it is 96. In the docs I found that the default reference value for PDF should be 72, which does not result in the 1.5 scale. It also contradicts the DPI of 144 that I found using tinkering: https://gist.github.com/oripka/36f4589df0fa72ed5758261c0a870264


```
    const dpi = await page.evaluate(() => {
      const div = document.createElement('div');
      div.style.position = 'absolute';
      div.style.top = 0;
      div.style.left = 0;
      div.style.width = '1in';
      div.style.height = '1in';
      document.body.appendChild(div);
      const height = div.offsetHeight;
      div.remove();
      return height;
    });
```

Maybe we should provide a sane default, test on Windows, Linux and macOS, in case there are any differences and make it a command line option.

PR #1539, opened 2024-04-16
feat: setting up hookable

antfu:
- [x] Hooks for server-side
- [ ] Hooks for client-side
- [ ] Setup files to get the app instances
- [ ] Docs
- [ ] Hooks for all setup equivalent capability


KermanX:
What's the status of this PR?

PR #1546, opened 2024-04-17
fix: disable code interop in inline code, fix #1540

### LINKED ISSUE #1540: Mustache Syntax is always interpreted
#524 seems to reoccur, I guess probably related to the MagicString rewrite

antfu:
I am not sure why it was disabled before. But I think it would make sense to have it as the better default.


antfu:
Oh I see, it was for Twoslash - I will find another solution

PR #1557, opened 2024-04-28
feat: rewrite VSCode extension

KermanX:
This PR rewrites the original VSCode extension using `@vue/runtime-core` for state management.

### Features

- Multiple project support
- Multiple markdown support (slides imported by `src:` option)
- Drag and drop slides in the slides tree
- Better preview experience
- Compat mode for `@slidev/cli` older than this PR
- Everything is reactive, even configuration
- ... (some UX details)

### Todos

- [x] Version & publish scripts
- [x] Improve stability for sync between slides tree, preview webview and cursor position
- [x] Whatever I have tried, `tsup` always bundles `vue.runtime-core.prod.cjs` instead of `.bundler.js`
- [ ] Docs
- [x] Preview panel
- [x] Multiple entry support
- [x] Icon looks a little bit small

#### How to debug?

1. `pnpm dev` on monorepo root.
2. Press `F5` in VSCode to launch the extension development host.


antfu:
The versioning would be a little bit tricky as VS Code Marketplace does not support `-beta.x` suffix in the version. I am still not sure what would be the best way to have VS Code extension inside a single-versioned monorepo

KermanX:
> The versioning would be a little bit tricky as VS Code Marketplace does not support `-beta.x` suffix in the version. I am still not sure what would be the best way to have VS Code extension inside a single-versioned monorepo

The problem of placing VSCode extension in another repo is that `@slidev/parser` is often updated when implementing a new feature in VSCode extension.

Could we modify the script and CI to manage VSCode extension's version and publish it separately?

antfu:
> Could we modify the script and CI to manage VSCode extension's version and publish it separately?

Sure we could. Than maybe we share with the same minor version, where patch versions are decoupled. In this way we could do 0.59.0 for the first version (so no beta).

KermanX:
> > Could we modify the script and CI to manage VSCode extension's version and publish it separately?
> 
> Sure we could. Than maybe we share with the same minor version, where patch versions are decoupled. In this way we could do 0.59.0 for the first version (so no beta).

Maybe we even needn't to make any connection between Slidev's main version with VSCode extension's. This is because VSCode extension doesn't use NPM to get dependencies - all dependencies are bundled when publishing.

PR #1587, opened 2024-05-11
fix: uniform client features enablement

KermanX:
This PR uniforms the feature enablement, and fixes that disabled features were displayed on the context menu

### DIFF
--- a/packages/client/internals/Controls.vue
+++ b/packages/client/internals/Controls.vue
@@ -3,20 +3,22 @@ import { shallowRef } from 'vue'
 import { showInfoDialog, showRecordingDialog } from '../state'
 import { configs } from '../env'
 import { useNav } from '../composables/useNav'
+import { useFeatures } from '../composables/useFeatures'
 import QuickOverview from './QuickOverview.vue'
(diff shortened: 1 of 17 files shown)


packages/client/internals/Controls.vue
```diff
@@ -3,20 +3,22 @@ import { shallowRef } from 'vue'
...
+const features = useFeatures()
+const drawingEnabled = features.drawings && !configs.drawings.presenterOnly && !isEmbedded.value
 const DrawingControls = shallowRef<any>()
 if (drawingEnabled)
   import('../internals/DrawingControls.vue').then(v => DrawingControls.value = v.default)
 
 const WebCamera = shallowRef<any>()
 const RecordingDialog = shallowRef<any>()
-if (__SLIDEV_FEATURE_RECORD__) {
+if (features.record) {
```
antfu, line 21: The reason we use the flag and static replacement is for tree-shaking. When the feature is disabled, the corresponding module should not be shipped to user's client bundle at all because Rollup will remove `if (false) {}`

Is the new approach still behaviours like that?

PR #1605, opened 2024-05-21
fix: should restart server on feature change

KermanX:
Previously, when `data.features` changed, the browser page will get a full reload but the server isn't started. This caused some problems. For example, the Katex CSS is not loaded when user adds the first Katex code block.

This PR restarts the server when the `katex` and `monaco` features become enabled.

### DIFF
--- a/packages/slidev/node/vite/loaders.ts
+++ b/packages/slidev/node/vite/loaders.ts
@@ -47,22 +47,6 @@ export function getBodyJson(req: Connect.IncomingMessage) {
   })
 }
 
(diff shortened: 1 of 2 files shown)


packages/slidev/node/vite/loaders.ts
```diff
@@ -47,22 +47,6 @@ export function getBodyJson(req: Connect.IncomingMessage) {
   })
 }
 
-export function sendHmrReload(server: ViteDevServer, modules: ModuleNode[]) {
```
KermanX, line 50: This function isn't used.

antfu (approved):

PR #1610, opened 2024-05-24
Add virtual scroll for overview and quick overview

jakobkordez:
Fixes slow loading and scrolling for large presentations

### DIFF
--- a/packages/client/internals/QuickOverview.vue
+++ b/packages/client/internals/QuickOverview.vue
@@ -1,11 +1,12 @@
 <script setup lang="ts">
-import { useEventListener } from '@vueuse/core'
+import { useEventListener, useVirtualList } from '@vueuse/core'
 import { computed, ref, watchEffect } from 'vue'
 import { breakpoints, showOverview, windowSize } from '../state'
 import { currentOverviewPage, overviewRowCount } from '../logic/overview'
 import { createFixedClicks } from '../composables/useClicks'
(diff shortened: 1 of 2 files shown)


KermanX:
Thanks for your contribution!

There is a problem that the height of the rows is calculated dynamically, but the `itemHeight` given to `useVirtualList` is a static value. I am not very sure on this.

KermanX (approved):

KermanX:
> There is a problem that the height of the rows is calculated dynamically, but the `itemHeight` given to `useVirtualList` is a static value. I am not very sure on this.

Fixed.

jakobkordez:
Don't merge yet, fixing something

jakobkordez:
Ok, that's all

KermanX (approved):

antfu:
I am not very sure about this - after this PR, every time you open up the overview, every slide gets re-initiated, which they are supposed to be cached.

If perf/memory is really a concern, I'd suggest we take a screenshot of the final slide and place the overview with that image (with an option to opt-in/out) so we don't duplicate slides rendering and would be a lot more performant than virtual scroll.

jakobkordez:
If you need a large presentation for testing try: https://gitlab.vegova.si/rkv/prosojnice

jakobkordez:
@antfu Are you sure the slides are not getting cached? It seems to me they are: the loading component only shows the first time the page is opened

PR #1629, opened 2024-05-30
fix: Updated types of PreparserSetup

jhnns:
It seems like the types of the `PreparserSetup` are outdated. At least they don't match what's described at https://sli.dev/custom/config-parser#preparser-extensions :)

### DIFF
--- a/packages/types/src/types.ts
+++ b/packages/types/src/types.ts
@@ -114,7 +114,7 @@ export interface SlidevData {
 }
 
 export interface SlidevPreparserExtension {
-  name: string
+  name?: string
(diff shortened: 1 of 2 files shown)


packages/types/src/types.ts
```diff
@@ -114,7 +114,7 @@ export interface SlidevData {
 }
 
 export interface SlidevPreparserExtension {
-  name: string
+  name?: string
```
jhnns, line 117: I wasn't sure about this one, but I think `name` is optional. At least there's no type error and the example doesn't use a `name` either.

KermanX (approved):
Thanks for your contribution!

PR #1630, opened 2024-05-31
fix: Goto layout

jakobkordez:
Fixes horizontal overflow (horizontal scrollbar on chrome)
Fixes layout for large slide numbers


KermanX (approved):
Thanks for your contribution!

Issue #1227, opened 2023-12-23
Supporting Synchronization for Slidev Deployments in Production Environments


li1553770945:
**Is your feature request related to a problem? Please describe.**

In the current implementation, the Slidev development environment initiated via `yarn dev` synchronizes page transitions, mouse clicks, and drawing actions in the presenter view across multiple users. This synchronization feature is crucial for collaborative usage.

However, deploying Slidev on a cloud server for multi-user access presents challenges. The development mode (`yarn dev`) is not suitable for production deployment. Conversely, deploying a static build using `yarn build` results in the loss of synchronization capabilities, as static pages inherently do not support dynamic backend interactions.

One of my thoughts is, would it be possible to support  **listener functions for these operations**, **as well as providing functions to be able to manually set page switching (currently implemented) and painting**? This way, users can use these functions to manually implement their own synchronization mechanisms.

**Describe the solution you'd like**

**Event Listener Integration**: Enable developers to define custom listener functions for slide transitions, mouse clicks, and drawing actions. These functions should trigger upon user interactions, such as page changes or drawing modifications.

**Serialization Support**:Enable developers to retrieve the current `currentPage` and `clicks`( these have now been implemented), along with the serialized drawing data.This would facilitate the transfer of drawing data across a network.

**Control and Animation Functions**: Introduce methods to programmatically set the slide `currentPage`, `clicks`(**Toggle animation is also supported**). Allow developers to apply all current drawings using serialized data (obtained in the previous step and transmitted over the network).

**Describe alternatives you've considered**

Currently, I am utilizing Vue's watch function to monitor changes in `currentPage` and `clicks`. However, this approach lacks efficiency and elegance. Moreover, I am uncertain about the best method to obtain a serialized form of the drawings for network transmission.

fullmetal-fred:
Just wanted to echo this one. I'm running a multi-day class out of my slidev project...students are expected to pull up the presentation locally and follow along. `slidev --remote` works fine for the sync experience, so I've built a container around it and deployed it to my cluster for now...but the page loads can be really wonky...often times I have to hard refresh my browser window in order for the presentation to come up after changes. Sometimes it just crashes on first load then refreshes and works fine. This experience, while fine for dev, isn't fine for less technical users in a prod use case. It would be great if static site generation "just worked" with sync if enabled in frontmatter.

KermanX:
I agree this feature is quite useful.

[...]

PR #1648, opened 2024-06-03
docs: add docker option to static hosting page

jakobkordez:
Adds a better and clearer guide to hosting static slides in a Docker container than is described in https://sli.dev/guide/install#install-on-docker

### DIFF
--- a/docs/guide/hosting.md
+++ b/docs/guide/hosting.md
@@ -195,3 +195,65 @@ jobs:
 
 - In your repository, go to Settings>Pages. Under "Build and deployment", select "GitHub Actions".
 - Finally, after all workflows are executed, a link to the slides should appear under Settings>Pages.
+
+### Docker
+
+To deploy your slides in a Docker container:
+
+- Create a `.dockerignore` file in your project root with the following content.
+
+```dockerignore
+node_modules
+dist
+```
+
+- Create a `.htaccess` file in your project root with the following content.
+
+```apache
+RewriteEngine On
+RewriteBase /
+RewriteRule ^index.html$ - [L]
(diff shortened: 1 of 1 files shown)


KermanX:
Thanks for your contribution!

I know almost nothing about Docker. Could you please also have a look at https://github.com/slidevjs/container?

jakobkordez:
What would the purpose of `slidevjs/container` be?
Currently `slidevjs/container` is a bad solution for statically hosting slides, it runs with a node image and never runs `slidev build`, which means it will never actually build a static website but will run a development server with node.
My commit uses a node image to run `slidev build` then copies only the `dist` folder to an `apache` image for hosting the static 

My commit only replaces https://sli.dev/guide/install#build-hostable-spa-single-page-application so maybe remove only that?

KermanX:
> What would the purpose of `slidevjs/container` be?
> Currently `slidevjs/container` is a bad solution for statically hosting slides, it runs with a node image and never runs `slidev build`, which means it will never actually build a static website but will run a development server with node.

The `slidevjs/container` runs a dev server because the built version doesn't support navigation sync.(#1227)

After #1227 is done, I think there can have a standalone navigation sync server as a Docker image.

KermanX:
There seem to be so many ways to deploy Slidev via docker. AKAIK, there are:

- [tangramor/slidev_docker](https://github.com/tangramor/slidev_docker) - Run Slidev in dev mode in a container
- [slidevjs/container](https://github.com/slidevjs/container) - The same
- [In the old docs: installation#host-via-docker](https://sli.dev/guide/install#host-via-docker) - Build on local machine. Docker only serves as an HTTP server
- [This PR](#) - Build and serve in Docker

PR #1661, opened 2024-06-11
docs: grammatical updates

QuinnyPig:
A bunch of non-substantive updates to the documentation to improve readability and flow.

PR #1643, opened 2024-06-01
docs: typos

thetestlabs:
Minor updates to the documentation at the root of the repository. This PR resolves typos and/or corrects small grammatical issues.

PR #1662, opened 2024-06-11
docs: typos

jakobkordez:
Fixes bad grammar and typos from #1661 and #1643


antfu:
Thank you - for the record - there are no bad PRs. People all make mistakes; some are typos, some are bugs, and none of them are big deals, and that's why we are doing OSS to help each other produce better software. Positive communications would help the community grow and sustain in the long run.

PR #1676, opened 2024-06-15
feat: optional wake lock

### LINKED ISSUE #1552: Keep slide show tab alive—Safari locks it in incognito mode
**Is your feature request related to a problem?**
Yes, when I open the slide show in full screen in Safari Private Browsing, the browser locks the window after ~10 minutes. This means that it also exits from the full screen view. ([live example](https://www.youtube.com/live/vkPkjzaun4o?si=duXNaV-4V1Q6T6Jo&t=4452))

Somehow keep the tab alive (like YouTube does) to avoid Safari locking it.

Disabling the _Require Touch ID to view locked tabs_ option in Safari. While this is a good workaround, it's not as convenient.

KermanX:
resolve #1552.

### DIFF
--- a/docs/custom/index.md
+++ b/docs/custom/index.md
@@ -59,6 +59,8 @@ selectable: true
 record: dev
 # enable Slidev's context menu, can be boolean, 'dev' or 'build'
 contextMenu: true
+# enable wake lock, can be boolean, 'dev' or 'build'
+wakeLock: true
 
 # force color schema for the slides, can be 'auto', 'light', or 'dark'
 colorSchema: auto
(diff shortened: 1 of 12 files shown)


antfu:
Awesome idea!

antfu:
I think we could have it enabled by default (when the app is on the front). With a client button and context menu to toggle it. And the frontmatter option can disable this feature completely

KermanX:
> I think we could have it enabled by default (when the app is on the front). With a client button and context menu to toggle it. And the frontmatter option can disable this feature completely.
> 
> In client, the state can be `Keep awake when slides in front` `Always keep awake` `Do not keep awake`

Added this option to the settings popover:
image

If the tab is not on the front, the wake lock will become effectless automatically by the browser.

antfu (approved):

PR #1677, opened 2024-06-16
fix(vscode): should pass entry file name to CLI

ras0q:
This PR fixes an issue in which the "Start Dev Server" command executed with the VSCode extension can build only default "slides.md".
For building custom markdown files, the filename should be explicitly specified.


KermanX (approved):
Thanks for you contribution!

PR #1692, opened 2024-06-20
fix: remove filepaths when building

### LINKED ISSUE #1690: problems about emitted files
_**Originally posted by @rwnobrega in https://github.com/slidevjs/slidev/issues/1669#issuecomment-2178938722**_

As of version `0.49.11`, most files are now free from local directory paths. However, the following files still expose local paths:

```
dist/assets/index-[hash].js
dist/index.html
dist/404.html
```

Additionally, as a side effect of the patch, some filenames now have unusual suffixes before the hash, which did not appear before. For example:

```
Arrow.vue_vue_type_script_setup_true_lang-[hash].js
```

instead of just:

```
Arrow-[hash].js
```

KermanX:
fix #1690. Removes file paths when building.

#1690 also reports some problems with chunk file names, but I can't figure out a good solution yet - Improvements to the chunking may need to be delayed.

### DIFF
--- a/packages/client/pages/overview.vue
+++ b/packages/client/pages/overview.vue
@@ -165,7 +165,7 @@ onMounted(() => {
             <carbon:presentation-file />
           </IconButton>
           <IconButton
-            v-if="route.meta?.slide"
+            v-if="__DEV__ && route.meta?.slide"
             class="mr--3 op0 group-hover:op80"
             title="Open in editor"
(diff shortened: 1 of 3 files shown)


rwnobrega:
> #1690 also reports some problems with chunk file names, but I can't figure out a good solution yet - Improvements to the chunking may need to be delayed.

FYI: There is an upstream issue on this: https://github.com/vitejs/vite-plugin-vue/issues/19. I hope this is useful.

PR #1698, opened 2024-06-24
feat: replace CodeMirror in side editor with Shiki-based textarea

### LINKED ISSUE #1391: RFC: Migrate CodeMirror to Monaco
Currently, we use CoreMirror for the in-app editor, while we support Monaco in the sides. With #1330 solving the major problem we had in Monaco, it makes it less optimal to maintain two different text editors. Now, with the Shiki Monaco integration, we could have a much better syntax highlight for the markdown content as well.

Consider the in-app editor only has a single instance and it's dev only, performance shouldn't a major concern.

This is also related to #1390

KermanX:
A breaking change because old custom styles may no longer work.

### Pros
- Reduce bundle size
- Highlighting is more accurate
- No CJS problems introduced by CodeMirror
- Load faster
- Easier to maintain

### Cons
- Undo/redo functionality becomes weak
- I have to use `word-break: break-all`, otherwise the text inside `<pre shiki>` and `<textarea>` will be mismatched, because `<span>`s in `<pre shiki>` affects the word breaking.

### Why not use monaco-editor
- If monaco-editor is used in the side editor, it has to be loaded in dev, even the user doesn't use `{monaco-*}` features
- Kind of too complex. Most of the features aren't useful in the side editor.

Resolve #1391.


packages/slidev/node/virtual/shiki.ts
```diff
@@ -70,6 +70,20 @@ export const templateShiki: VirtualModuleTemplate = {
...
+      'let highlight',
+      'export async function getHighlighter() {',
+      '  if (highlight) return highlight',
+      '  const highlighter = await shiki',
+      '  highlight = (code, lang, options) => highlighter.codeToHtml(code, {',
+      '    lang,',
+      `    theme: ${JSON.stringify(themeOptionsNames.theme)},`,
+      `    themes: ${JSON.stringify(themeOptionsNames.themes)},`,
+      '    defaultColor: false,',
+      '    ...options,',
```
KermanX, line 83: Moved here from https://github.com/slidevjs/slidev/blob/4707949b6612a0f1cc60f8c969ea7dadf4727862/packages/client/setup/code-runners.ts#L20-L28.

It now uses Shiki's dual theme feature. I am not sure why it wasn't used.

PR #1708, opened 2024-06-28
fix: hmr experience

KermanX:
1. fix: Circular dependency caused flash during HMR
2. perf: Update preload strategy. Previously, it depended on `QuickOverview`, now its logic is written in `SlidesShow`.
3. fix: no longer pre-mount `QuickOverview` (i.e. `v-if="showOverview"` instead of `v-show="showOverview"`)
4. fix: the last "Preview Next" in presenter mode disappears
5. fix: `clicksContext.onMounted/onUnmounted` should be called correctly.

Blocked by https://github.com/vuejs/core/pull/11248


antfu:
Do we need to wait until the next Vue patch?

PR #1718, opened 2024-06-29
feat: handle loader error

KermanX:
Preview: (previously the annotations just didn't update anymore without showing any error)

image

Also include:
- Some package.json fix
- Use sync fs operations to speed up


packages/parser/src/fs.ts
```diff
@@ -110,8 +129,8 @@ export async function load(userRoot: string, filepath: string, content?: string,
   }
 }
 
-export async function save(markdown: SlidevMarkdown) {
+export function save(markdown: SlidevMarkdown) {
```
antfu, line 132: I would prefer to keep this function async, in case we need async operation in the future.

antfu (approved):

PR #1720, opened 2024-06-30
docs: only allow UnoCSS as the CSS engine

KermanX:
Previously, the CSS engine could be set to none, but Slidev itself is depending on UnoCSS.

This PR also clarifies that Tailwind classes can be used out-of-box. This is for people who are afraid of learning a new CSS framework.


antfu:
Got a smoke error, otherwise LGTM

PR #1731, opened 2024-07-04
ci: upgrade `pnpm/action-setup`

KermanX:
ref https://github.com/pnpm/action-setup/issues/135

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -33,9 +33,6 @@ jobs:
       - name: Setup
         run: npm i -g @antfu/ni
 
-      - name: Setup PNPM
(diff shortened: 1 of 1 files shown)


.github/workflows/smoke.yml
```diff
@@ -33,9 +33,6 @@ jobs:
       - name: Setup
         run: npm i -g @antfu/ni
 
-      - name: Setup PNPM
-        uses: pnpm/action-setup@v2
```
antfu, line 37: Why do we remove this?

KermanX: This seems to be unused. Also, test.ci and release.ci don't have this.

PR #1744, opened 2024-07-09
feat: support relative import inside slides

KermanX:
This PR is based on #1708. It will be much smaller after merging that one.

Previously, it wasn't possible to import assets via relative paths. For example, ``. This was because the slides contents were loaded as `/@slidev/slide/1.md`.

This PR loads the slides in `path/to/slides.md` as `path/to/slides.md__slidev_x.md`, where `x` is the slide no in the presentation, instead of the slide no in that Markdown file. And `/@slidev/slide/x.md` works as a facade which re-exports `path/to/slides.md__slidev_x.md`. The same for frontmatter.

The reason why not use urls like `path/to/slides.md?slidev=x.md` is that this will conflict with Vue's logic. The Vue plugin doesn't preserve the query of the URL. Also, Vue plugin's `include` option only matches the path part without query string.

This PR also splits `vite/loaders.ts` into multiple files - each Vite plugin has its own file. And removed most of the usages of `enforce`/`order`.


antfu:
Thanks for working on it! You are amazing!

antfu (approved):
I am ok with the direction, I'll let you handle the merge and release

PR #1750, opened 2024-07-11
chore: use pnpm catalog

btea:
https://pnpm.io/catalogs


KermanX:
Does https://github.com/antfu-collective/taze support PNPM catalog?

btea:
I checked the code and it seems that it is not supported at present.

btea:
@KermanX I created a PR. https://github.com/antfu-collective/taze/pull/123

antfu:
We you like to help solving the conflicts, or redo the PR? Thanks

KermanX:
The problem is that our smoke test CI needs to pack the packages. Not sure why `pnpm pack` doesn't resolve `catelog:` protocol.

btea:
There were a lot of conflicts, so I created a new PR #1820.

btea:
> The problem is that our smoke test CI needs to pack the packages. Not sure why `pnpm pack` doesn't resolve `catelog:` protocol.

According to the error message, it seems that npm was used to install dependencies, but npm does not support the `catalog` protocol.

PR #1761, opened 2024-07-18
refactor: re-organize plugins and setup loaders

KermanX:
- Start-up time (from <kbd>enter</kbd> pressed to the URL shown in terminal) of demo slides is reduced by about 0.8s
- More maintainable codebase
- Allow to provide options for all 3rd party Vite plugins that are used internally


antfu (approved):

antfu:
Nice refactoring!

PR #1797, opened 2024-08-03
feat: add CLI option to allow transparent pngs

flwst:
PNGs are generated by taking a screenshot of the page as if it were running in a browser using Playwright. Currently, PNG exports will never be transparent and will always have the default, white, browser background regardless of if the user has removed all backgrounds using css. Playwright has an `omitBackground` property that can be added to the `.screenshot` method to remove this final browser background.

https://playwright.dev/docs/api/class-page#page-screenshot-option-omit-background

This PR adds the `--omit-background` CLI option to control the visibility of the default browser background and allow for the possibility of transparent png exports. It also adds docs for this feature

Users will still need to apply additional css styling to remove the css backgrounds from their slides in order to reveal the transparency in their exports.

### DIFF
--- a/docs/builtin/cli.md
+++ b/docs/builtin/cli.md
@@ -74,6 +74,7 @@ Options:
 - `--dark` (`boolean`, default: `false`): export as dark theme.
 - `--with-clicks`, `-c` (`boolean`, default: `false`): export pages for every click animation (see https://sli.dev/guide/animations.html#click-animation).
 - `--theme`, `-t` (`string`): override theme.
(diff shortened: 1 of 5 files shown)


KermanX (approved):
Thanks!

I think it's better to name this option `--omit-background` for clarity.

antfu (approved):

antfu:
Yeah `--omit-background` or `--no-background` sounds better to me as well

KermanX (approved):

PR #1800, opened 2024-08-05
feat: run Slidev in browser

KermanX:
[WIP] This PR tries to run Slidev in the browser, with a better editor. I think the most commonly used functions work in the browser, but some don't. In this PR, changes to existing code will be very minimal, so the Node.js version will always be workable.

Although Slidev can run in StackBlitz, it is slow regarding startup time and operational efficiency.

### DIFF
--- a/packages/client/logic/utils.ts
+++ b/packages/client/logic/utils.ts
@@ -1,4 +1,4 @@
-import { parseRangeString } from '@slidev/parser/core'
+import { parseRangeString } from '@slidev/parser/utils'
 import { useTimestamp } from '@vueuse/core'
 import { computed, ref } from 'vue'
 
--- a/packages/parser/package.json
+++ b/packages/parser/package.json
@@ -23,6 +23,10 @@
     "./fs": {
       "types": "./dist/fs.d.mts",
       "import": "./dist/fs.mjs"
+    },
+    "./utils": {
+      "types": "./dist/utils.d.mts",
+      "import": "./dist/utils.mjs"
     }
   },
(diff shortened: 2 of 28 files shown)


cnguyen-de:
If the user can download an exported pdf version from the live version, this feature would be a killer feature for slidev.

cnguyen-de:
I tried out the PR and added some code on my local branch to test if you can add new slide with custom template other than default one.

```
packages/web/src/virtual/nav-controls.vue

const selectedTemplate = ref('default')

function add() {
  slidesSource.push({
    frontmatter: { layout: selectedTemplate.value },
    content: `# ${slidesSource.length + 1}`,
    note: '',
  })

...

// should be dynamically generated based on the current theme
const templates = [{
  value: 'default',
}, {
  value: 'fact',
}, {
  value: 'section',
}, {
  value: 'quote',
}]
</script>

<template>
  <MenuButton>
    <template #button>
      <IconButton class="slidev-icon" title="Add Slide">
        <carbon:add-large />
      </IconButton>
    </template>

    <template #menu>
      <SelectList v-model="selectedTemplate" title="Template" :items="templates" @click="add()" />
    </template>
  </MenuButton>
...
```

However whenever a new slide is added, the layout being shown is still the layout of the last slide. When inspecting the value in devtools,  the value of frontmatter is correct. Upon creating new slide, the layout is applied. But same issue happens again, so basically the frontmatter config is delayed by one slide. I don't know all the internals yet to find a fix so I'm writing here to report the bug.

PR #1838, opened 2024-08-24
fix: should resolve icons from all roots

KermanX:


packages/slidev/node/options.ts
```diff
@@ -64,7 +64,7 @@ export async function resolveOptions(
     roots,
   }
 
-  return {
+  return globalThis.__SLIDEV_OPTIONS__ = {
```
antfu, line 67: Do we have to use global variables to do this tho?

KermanX: Removed the usage of `globalThis` by extracting `presetIcons` from `client/uno.config.ts` to `node/setups/unocss.ts`

PR #1842, opened 2024-08-26
docs: fix indicator for current page and hydration mismatch

warmthsea:
#### Before:

screenshots

#### After:

screenshots


KermanX:
`<LinkInline link="guide/" />` is displayed as `Why Slidev` instead of `Getting Started` (although the target URL is correct)

KermanX (approved):

Issue #855, opened 2023-02-01
Presenter sync in GitHub pages


emersonbottero:
Is it possible to use remote while I use /presenter in the GitHub page and n given number of people can follow along in the default GitHub page.

Would be one hell of a feature... 😆

tonai:
For sure that woul be great, but in fact I don't think that remote synchronization is possible without using a server.
Event technologies like WebRTC needs to send first some "handshake", from one client to another one, before establishing a peer to peer connetion.

In development the synchronization uses [vite-plugin-vue-server-ref](https://github.com/antfu/vite-plugin-vue-server-ref) that make it possible using your local vite server.
But once it is builded and deployed there is no server anymore...

One solution would be to have a fallback to some WebSocket or SSE server for example for server refs.
But who will host (and pay for) that server ?
In free plan, Sass solutions are limited, so having some kind of central slidev server won't work.

So maybe the only real solution is, for every slidev user, that want remote synchronization in build mode, to deploy their own server.

I created a server I use for one addon I created that adds a poll component to slidev.
* the addon: https://github.com/Smile-SA/slidev-component-poll
* the server: https://github.com/Smile-SA/slidev-poll-server

It uses WebSocket or Server Sent Events to synchronize the clients, and it works when the slides are statically builded and deployed.

Maybe we can do something that is close to that, in a more generic way (it would be great if it can handle any kind of server ref for example).
Something that can be easily deployable and pluggable.
In that case using a WebRTC can be interesting because the server will only manage the "handshake" between the clients and the rest of the logic will be managed on the client side.

Now I think there is still some questions:
* where can someone easily deploy a node server at no cost (AWS lambda maybe ?)
* how do we package the server so it is easily usable ?
* how do we manage security ?

barais:
@tonai 

I just provide an initial draft of an addon for that purpose. It is based on a fork version of your poll component and poll server. 
If it could be useful to someone else, do not hesitate. 

- [x] slide navigation and click synchro
- [x] optional cursor synchronization  
- [ ] TODO: drawing synchronization.  

https://github.com/barais/slidev-addon-ws-syncho
https://github.com/barais/slidev-slide-synchro-server

It works with github page and the ws server behind nxing proxy and lets encrypt certificate. 

Do not hesitate to criticize/comment.

tonai:
For info I created a PR to centralize the handling of state synchronization here: https://github.com/slidevjs/slidev/pull/1854
[...]

PR #1854, opened 2024-09-06
feat: create extendable sync states

tonai:
Extends `createSyncState` to allow custom external synchronization.

the goal is to solve the problem with synchronization using static builds:
* https://github.com/slidevjs/slidev/issues/855
* https://github.com/slidevjs/slidev/issues/1227

This can be a quick solution before migrating on Nuxt.

I am already working on an addon that will allow synchronization of shared and drawings states (and possibly others).
* The addon: https://github.com/Smile-SA/slidev-addon-sync
* The SSE and WS server that can be used for the synchronization: https://github.com/Smile-SA/slidev-sync-server

### DIFF
--- a/packages/client/index.ts
+++ b/packages/client/index.ts
@@ -9,5 +9,11 @@ export { useDrawings } from './composables/useDrawings'
 export { useDarkMode } from './composables/useDarkMode'
 export { useIsSlideActive, onSlideEnter, onSlideLeave } from './logic/slides'
 
+export { createSyncState, disableSlidevSync, addSyncMethod } from './state/syncState'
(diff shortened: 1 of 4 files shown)


tonai:
For information I also updated my slidev poll addon https://github.com/Smile-SA/slidev-component-poll that uses `createSyncState` to now let https://github.com/Smile-SA/slidev-addon-sync handle the synchronization for static builds and it works just fine.
I am just waiting for this PR to be merged and a new release to be published to publish both `slidev-component-poll` and `slidev-addon-sync` with the correct slidev `peerDependencies`

antfu (approved):

PR #1869, opened 2024-09-18
fix: typo on pnpm create command

stefanobartoletti:
Fixes a typo in the docs where `pnpm create` was incorrectly referred as `pnpm init`


KermanX (approved):
Thanks!

PR #1891, opened 2024-10-07
docs: fix sourceLink of the Rust talk

hongquan:
The old URL only works with Git tool. Fix to make it accessible via browser.

### DIFF
--- a/docs/.vitepress/showcases.ts
+++ b/docs/.vitepress/showcases.ts
@@ -222,7 +222,7 @@ export const showcases: ShowCaseInfo[] = [
     },
     at: 'FOSSASIA Summit 2024',
     slidesLink: 'https://talk.quan.hoabinh.vn/rust-error-handling-ease-web-dev/',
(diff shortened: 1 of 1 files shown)


KermanX:
The new link shows "Having trouble showing that directory". Is this intentional?

image

docs/.vitepress/showcases.ts
```diff
@@ -222,7 +222,7 @@ export const showcases: ShowCaseInfo[] = [
     },
     at: 'FOSSASIA Summit 2024',
     slidesLink: 'https://talk.quan.hoabinh.vn/rust-error-handling-ease-web-dev/',
-    sourceLink: 'https://hongquan@bitbucket.org/hongquan/rust-error-handling-ease-web-dev.git',
```
antfu, line 225: Seems like it should be `https://bitbucket.org/hongquan/rust-error-handling-ease-web-dev/src/main/`

I merged it too fast 🫠

hongquan:
I'm sorry. The correct URL is "...web-dev/" (with trailing slash). I thought that BitBucket automatically redirects from non-slashed to with-slashed, but I was wrong.

PR #1895, opened 2024-10-09
docs: add more community addons

tonai:


carlobellettini:
repo links are missing the `s` of `slidev` 6 out 7 times

for example

>    repo: 'https://github.com/Smile-SA/lidev-component-spotlight',

must be

>    repo: 'https://github.com/Smile-SA/slidev-component-spotlight',

docs/.vitepress/addons.ts
```diff
@@ -25,6 +25,76 @@ export const community: AddonInfo[] = [
...
+    name: 'slidev-addon-sync',
+    description: 'Sync component for Slidev static build that uses a SSE or WS server',
+    author: {
+      name: 'Tony Cabaye',
+      link: 'https://github.com/tonai',
+    },
+    repo: 'https://github.com/Smile-SA/slidev-addon-sync',
+  },
+  {
+    id: 'slidev-component-pager',
```
antfu, line 39: Might be better to have a components category then? What do you think @KermanX?

KermanX: How about adding some tags/keywords to each addon? For example, `Component`/`Intergration`/`Tool`

tonai: Just to add some context, the first add-ons I created were named `slidev-component-*` because I was working on them before we released the add-on feature in Slidev (released with v0.32 on May 20th).
The name "addon" was not official at that time, so I chose "component" and was too lazy to change the name later.
But yes, it can still be interesting to have tags :+1:

KermanX:
@tonai, I've just added some tags to the addons. However, I'm unsure about the accuracy of these tags. Could you please review them? 😊

tonai:
@KermanX Sorry for the delay.
The tags look fine.
And thanks for the hint I will check that.

Issue #1921, opened 2024-10-29
Static analysis of iframe src


tschaub:
**Is your feature request related to a problem? Please describe.**
I'm trying to use `iframe`s with relative `src` paths and was hoping that these would be handled in Slidev as they are in Vite (e.g. Vite serves the iframe sources and bundles scripts referenced in them).

I see that Slidev allows assets like images and videos to be ["imported" with relative paths](https://sli.dev/guide/faq#assets-handling).

**Describe the solution you'd like**
I was hoping it would be possible to have something like this in a slide:
```html
<iframe src="./example/basic.html"></iframe>
```

With that, I would like Vite to serve the `./example/basic.html` in development and bundle referenced scripts when doing a build.

**Describe alternatives you've considered**

I have considered running a second instance of Vite to host the iframe sources during development and to generate the build after building with Slidev. This sounds very awkward.

KermanX:
I found a workaround:

**vite.plugin.ts**:
```ts
import { defineConfig } from 'vite'

export default defineConfig({
  plugins: [],
  slidev: {
    vue: {
      template: {
        transformAssetUrls: {
          iframe: ['src'],

          // Default ones
          video: ['src', 'poster'],
          source: ['src'],
          img: ['src'],
          image: ['xlink:href', 'href'],
          use: ['xlink:href', 'href'],
        },
      },
    },
  },
})
```

**slides.md**:
```md
<iframe src="./basic.html?url"></iframe>
```

tschaub:
Thanks for looking into this, @KermanX.

I tried the workaround that you suggested, but I'm seeing the main `index.html` loaded (infinitely) instead of the iframe source.
<img width="640" alt="image" src="https://github.com/user-attachments/assets/5905e5b0-42df-4766-ad7a-99cb9c186cca">

Here is an effort to use the `transformAssetUrls` workaround: https://stackblitz.com/edit/slidev-iframe-plugin-ibx8tj?file=slides.md

The workaround I am currently using is to add the iframe sources to the `options.input` (so these are included when building) and then to add middleware in `configureServer` that overrides `response.end` (to avoid the dev server behavior of using the same `index.html` for all `*.html` requests). Not feeling great about that, but it does seem to be working.

Here is an example with the above hacky workaround: https://stackblitz.com/edit/slidev-iframe-plugin-upiwfu?file=vite.config.ts

PR #1926, opened 2024-11-01
fix: do not override user-provided html files

### LINKED ISSUE #1923: The slidev:config plugin doesn't allow Vite to serve other html files
**Describe the bug**

The `slidev:config` plugin configures the Vite dev server to load the built-in Vue index page for any request URLs ending with `.html`. This makes it so Vite cannot be used to serve additional pages that might be used as embedded iframes in a slideshow.

https://github.com/slidevjs/slidev/blob/3be1191199fc2d0814f60aed139c81a299477a79/packages/slidev/node/vite/extendConfig.ts#L189-L202

If a slideshow uses an `<iframe>` and the URL ends with `.html`, it will not be loaded by the dev server. A workaround is to use a URL that ends with `.htm` instead. See below for an example.

**Minimal reproduction**

1. Go to https://stackblitz.com/edit/slidev-iframe-plugin?file=slides.md,vite.config.ts
2. Advance the slideshow preview to see the working iframe on the second slide
3. Advance to the next slide to see the broken iframe on the third slide
4. See how it works when the iframe source URL ends with `.htm` instead of `.html`

**Environment**

- Slidev version: 0.50.0-beta.5
- Browser: Chrome 130
- OS: macOS 14.7

KermanX:
fix #1923

### DIFF
--- a/packages/slidev/node/vite/extendConfig.ts
+++ b/packages/slidev/node/vite/extendConfig.ts
@@ -190,7 +190,7 @@ export function createConfigPlugin(options: ResolvedSlidevOptions): Plugin {
       // serve our index.html after vite history fallback
       return () => {
         server.middlewares.use(async (req, res, next) => {
-          if (req.url!.endsWith('.html')) {
+          if (req.url === '/index.html') {
             res.setHeader('Content-Type', 'text/html')
             res.statusCode = 200
             res.end(options.utils.indexHtml)


tschaub:
Thanks for the quick fix, @KermanX!

Here is a working example using this fix: https://stackblitz.com/edit/wzuhnt?file=README.md

As mentioned there, a plugin is still needed to get `slidev build` to include the iframe sources. Perhaps that could be addressed with a fix for #1921.

PR #1928, opened 2024-11-01
feat: Timer stop

ken7253:
## Motivation

https://x.com/_yuheiy/status/1851861256283795513

Because he wanted a timer stop feature on slidev and I wanted the same feature.


KermanX (approved):
Thanks a lot!

In the latest commit I moved the composable to the `composables` directory, and updated the UI:

(When hovering the timer)
image

Issue #1663, opened 2024-06-12
Add support for lang attribute


jakobkordez:
**Describe the solution you'd like**
Add `lang` to frontmatter on a per deck or slide basis

**Describe alternatives you've considered**
An alternative would be to support setting the `lang` attribute in the `html` tag in `index.html`

PR #1948, opened 2024-11-15
feat: support lang attributes

ssssota:
close: #1663 

Apply lang attribute into page/index.html from frontmatter/headmatter.


antfu (approved):

KermanX (approved):
I am not sure if `htmlLang` will be a better name because this attribute is not very commonly used, while `lang` is a quite general name.

antfu:
I think `lang` should be fine. As Slidev works as a webpage, while `lang` is a widely used attribute name in HTML.

PR #1951, opened 2024-11-18
fix: setup preparser correctly when no addons are used

oripka:
Currently the preparser is only called if `addons.length > 0`. If no addons are used the preparser does not work. This fixes that issue.


KermanX (approved):
Thanks a lot~

PR #1963, opened 2024-11-30
fix(vscode): compute source offsets correctly

KazariEX:


KermanX (approved):
We made the same fix at the same time😂

PR #1965, opened 2024-12-01
chore: improve demo slides source

KermanX:


packages/client/styles/index.css
```diff
@@ -40,6 +40,10 @@ html {
   pointer-events: none;
 }
 
+.slidev-layout a.slidev-icon-btn {
+  @apply border-none hover:border-none hover:text-white;
```
KermanX, line 44: I tried not to use `!important` here

demo/starter/slides.md
```diff
@@ -26,19 +26,16 @@ mdc: true
...
-    Press Space for next page <div class="i-carbon:arrow-right inline-block"/>
-  </span>
+<div @click="$slidev.nav.next" mt-12 py-1 hover:bg="white op-10">
+  Press Space for next page <carbon:arrow-right />
 </div>
 
-<div class="abs-br m-6 flex gap-2">
-  <button @click="$slidev.nav.openInEditor()" title="Open in Editor" class="text-xl slidev-icon-btn opacity-50 !border-none !hover:text-white">
-    <div class="i-carbon:edit" />
+<div abs-br m-6 text-xl>
```
antfu, line 33: The reason to fix class and attributify as to demonstrate both mode are valid. People that wasn't using UnoCSS before might get confused if we use valueless mode up front

Issue #1705, opened 2024-06-26
CSS injected by UnoCSS incomplete on first load


jgosmann:
**Describe the bug**

Often (but not always, I wasn't able to reliably reproduce the issue), the CSS injected by UnoCSS is incomplete and slides appear broken. Most commonly the nav controls are shown incorrectly. After the first load, requesting the `__uno.css` will return all required styles and thus, a reload fixes the issue.

**Minimal reproduction**

Unfortunately, I cannot provide a reliable minimal reproduction. But maybe the video recording below helps. This is what happens in there:

1. I have started the Slidev presentation, but not loaded it in the browser.
2. I fetch `__uno.css` in the terminal. Observe that the CSS content isn't very long.
3. I load the presentation in the browser. The nav controls are not displayed correctly because some of the CSS styles are missing.
4. I fetch `__uno.css` once more in the terminal. Observe that the CSS content is now considerably longer. Diffing the previous content with the content now also reveals major differences.
5. I reload the slides in the browser. Now everything is displayed correctly.

https://github.com/slidevjs/slidev/assets/850157/714268bb-8c9c-47e0-870a-097c7386aa50

**Environment**

- Slidev version: v0.49.12
- Browser: Firefox 127.0.2 (couldn't reproduce it in Chrome 126.0.6478.116)
- OS: macOS 14.5

jgosmann:
There might be an interaction with the theme (that I unfortunately cannot share). The theme used defines its own `uno.config.ts` and sets shortcuts and custom theme colors in there. With the `seriph` theme I couldn't reproduce it so far.

KermanX:
Similar to https://github.com/slidevjs/slidev/issues/1573#issuecomment-2173315470.

PR #1969, opened 2024-12-02
fix: add frequently missing classes to unocss safelist

KermanX:
A temporary workaround for #1705.

### DIFF
--- a/packages/client/uno.config.ts
+++ b/packages/client/uno.config.ts
@@ -13,6 +13,7 @@ export default defineConfig({
   safelist: [
     '!opacity-0',
     'prose',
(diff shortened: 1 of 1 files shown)


packages/client/uno.config.ts
```diff
@@ -13,6 +13,8 @@ export default defineConfig({
   safelist: [
     '!opacity-0',
     'prose',
+    'grid-rows-[1fr_max-content]',
```
antfu, line 16: We could add a comment above, so we know why we have them and could remove it in the future if we figured out why

antfu (approved):

Issue #1732, opened 2024-07-04
A mechanism to specify which clicks should be exported on a per-slide basis


DavidAntliff:
Apologies if I've missed an existing feature request for this, I did search for one and didn't find anything relevant.

When exporting, without `--with-clicks`, the docs say:

> By default, Slidev exports one page per slide with clicks animations disabled.

From what I can tell, this is (always?) the last "click" for each slide. This is not always the most useful representation of a slide and it is sometimes useful to export an earlier click on that slide.

With `--with-clicks`, *every* transition for *every* slide is produced in the PDF, which can add a lot of unnecessary slides. Such an export can be manually edited to remove unnecessary slides, but this is actually really tedious.

Also, it is sometimes useful to have a few of the click states rendered on selected slides, but not all of them (for example in a lengthy magic-move sequence). Additionally, it may be useful to have some slides optionally omitted from the export (perhaps due to sensitive data).

A concrete example is a before/after "refactoring" slide, where, say, some "bad" code is shown on the first click, and then magic-move is used over, say, 20 animated steps, to work towards the final click which shows the "good" version of the code, and in the final export perhaps only the first and last clicks are actually useful in a PDF.

I propose a way to specify, for a specific slide (perhaps in the YAML frontmatter for the slide), the click or clicks that will be rendered to PDF when `--with-clicks` is *not* provided.

For example:

By default, the current behaviour for each slide continues, but for this page that has a, say, `export: OnClicks:` entry:

```
---
export:
  forClicks: 1-3,*
---

# My Slide

<v-clicks>
* one
* two
* three
* four
* five
</v-clicks>

```

When export is run with `--with-clicks`, the PDF would contain pages only for click states 1, 2, 3 and the final slide state for this specific page.

The `forClicks` field could take individual click numbers (e.g `1`), ranges (`3-5`), sequences delimited by `,`, and the special `*` to indicate the final click on that page, whatever that might ultimately be.

This would also allow pages to be skipped from the export if the `forClicks:` value is empty/null.

----

Since the default behaviour would remain unchanged, this would be a purely opt-in feature, and not affect any existing presentations.

KerstinKeller:
I am just exporting my slides, and agree that this would be very helpful.

antfu:
I agree it makes sense and the technically side it should be possible. But I see the problem more like how we should design the syntax. Any proposals/ideas?

DavidAntliff:
> Any proposals/ideas?

Yes, see above...

[...]

PR #1972, opened 2024-12-03
feat: browser exporter

kermanx:
This PR introduces a new page on `/export` as an alternate to the `slidev export` CLI.

- A new "export" button is added under the "more options" menu in the navigation bar.
- When running `slidev`, the exporting URL will be shown.
- When running `slidev export`, a tip will be shown to inform the user to use the exporting UI.

> [!warning]
>
> Both PDF exporting and image capturing don't work well on Firefox. And I don't have a Safari to test.

### The Approach

- For PDF, use the browser's built-in "export to PDF" function.
- For PPTX/PNGs, the "Screen Capture API" is used.

### Screenshots (by 2024/12/5 11:55)

QQ_1733370981699

image

*On non-Chromium or chrome<94*:

image

### Todos

Maybe adding some UI for #1732?


antfu (approved):
Let's have it as experimental and ship it to iterates. I have some UI adjustment in mind but I think the feature wise is very solid already. Thanks a lot for working on it!

docs/custom/index.md
```diff
@@ -27,6 +27,8 @@ keywords: keyword1,keyword2
 
 # enable presenter mode, can be boolean, 'dev' or 'build'
 presenter: true
+# enable exporting UI, can be boolean, 'dev' or 'build'
+exportingUI: dev
```
antfu, line 31: I think just `export` would be good enough :)

KermanX: We already have an `export` field in the headmatter:

```yaml
# export options
# use export CLI options in camelCase format
# Learn more: https://sli.dev/guide/exporting.html
export:
  format: pdf
  timeout: 30000
  dark: false
  withClicks: false
  withToc: false
```

antfu: How about `export.browser`?

antfu: I guess that would be fine. We could update the docs mentioning about that. The name `exportingUI` feels a bit verbose and we don't usually have the concept of `UI`. Or I don't know if we could call it `exporter` to differentiate with the CLI export

antfu: Or `browserExport` or `browserExporter` to be more specific?

antfu:
Let's have it and iterate on the main. I am open for better naming

packages/slidev/node/cli.ts
```diff
@@ -443,8 +443,19 @@ cli.command(
...
       const options = await resolveOptions({ entry: entryFile, theme }, 'export')
+
+      if (options.data.config.browserExporter !== false && !warned) {
+        warned = true
+        console.log(lightCyan('[Slidev] Try the new browser exporter!'))
+        console.log(
+          lightCyan('You can use the browser exporter instead by starting the dev server as normal and visit'),
+          `${blue('localhost:')}${dim('<port>')}${blue('/export')}\n`,
+        )
+      }
```
kalvdans, line 457: This text is now spamming my terminal when I run in batch mode. Would you accept a PR that silences it in batch mode?

PR #1988, opened 2024-12-27
feat: use screen capture for slides snapshot

### LINKED ISSUE #1883: Snapshot overview not color mode aware
Using the latest v0.50.0-beta1 and trying the experimental snapshot overview i noticed. That switching color modes does not switch the color mode of the snapshots, see screenshots.

In the blurred background you see I switched to light mode but the preview is still in dark mode. This happens not only on the loading slides.

image

### LINKED ISSUE #1889: Quick overview opening slowly on larger presentations
Quick overview takes a few seconds to open on large presentations (400 slides).
This is a big problem when I want to quickly jump to a slide.

This problem occurs because a very large amount of DOM elements need to be added during opening.

This problem was fixed in #1610 but was never merged.

Is there a possibility of this being merged if I resolve the conflicts or should I just create a fork for my personal use?

antfu:
Use the same screenshot logic from #1972 by @KermanX, remove the previous `html-to-image` approach.

- fix #1889
- fix #1883

### DIFF
--- a/packages/client/constants.ts
+++ b/packages/client/constants.ts
@@ -83,5 +83,4 @@ export const HEADMATTER_FIELDS = [
   'mdc',
   'contextMenu',
   'wakeLock',
-  'overviewSnapshots',
 ]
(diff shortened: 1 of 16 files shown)


KermanX (approved):
How about capturing multiple slides simultaneously, which will make the process faster?

PR #1989, opened 2024-12-27
fix: hide sync controls in SPA mode

KermanX:
As title.

Also, is it possible to simplify the sync controls UI?

btw, I am confused about the new screen mirroring feature - Why does the presenter want to see his own tab/screen?


antfu:
SPA syncing would also work in the same browser via Boardcast API, is in some cases it would still be useful. Maybe we could have an option to disable it, or smartly detect how many tab instances are created.

Screen mirroring feature is for presentation on a same laptop with two monitor.

PR #2026, opened 2025-01-29
perf: use Shiki shorthand

KermanX:
This speeds up the demo slides' start-up time by almost 2s on my machine.

TODO:

`[vite] Internal server error: highlighted.indexOf is not a function`: Maybe solved by `markdown-it-async@2`. To upgrade to v2, a release of https://github.com/unplugin/unplugin-vue-markdown is needed.

### DIFF
--- a/packages/slidev/node/syntax/markdown-it/markdown-it-shiki.ts
+++ b/packages/slidev/node/syntax/markdown-it/markdown-it-shiki.ts
@@ -1,7 +1,7 @@
 import type { ResolvedSlidevOptions } from '@slidev/types'
 import type { ShikiTransformer } from 'shiki'
 import { isTruthy } from '@antfu/utils'
-import { fromHighlighter } from '@shikijs/markdown-it/core'
+import { fromAsyncCodeToHtml } from '@shikijs/markdown-it/async'
 import { escapeVueInCode } from '../transform/utils'
 
 export default async function MarkdownItShiki({ data: { config }, mode, utils }: ResolvedSlidevOptions) {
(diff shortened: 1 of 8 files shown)


packages/types/src/setups.ts
```diff
@@ -57,8 +57,7 @@ export type ShikiSetupReturn =
     & CodeOptionsThemes<BuiltinTheme>
     & CodeOptionsMeta
     & {
-      setup: (highlighter: Highlighter) => Awaitable<void>
```
KermanX, line 60: It seems that this function is never called. And I don't know how to support it because there's no longer a `Highlighter`

antfu: We can get the highlighter with:

```ts
  const shiki = createSingletonShorthands<string, string>(createHighlighter)
  const highlighter = await shiki.getSingletonHighlighter()
```

But I don't mind removing it if no one uses it

packages/slidev/node/syntax/markdown-it/markdown-it-shiki.ts
```diff
@@ -1,21 +1,31 @@
...
-  const transformers = [
-    ...utils.shikiOptions.transformers || [],
-    (config.twoslash === true || config.twoslash === mode)
-    && (await import('@shikijs/vitepress-twoslash')).transformerTwoslash({
+export default async function MarkdownItShiki({ data: { config }, mode, utils: { shiki, shikiOptions } }: ResolvedSlidevOptions) {
+  async function getTwoslashTransformer() {
+    const [,,{ transformerTwoslash }] = await Promise.all([
+      // trigger the shiki to load the langs
+      shiki.codeToHast('', { lang: 'js', ...shikiOptions }),
+      shiki.codeToHast('', { lang: 'ts', ...shikiOptions }),
```
KermanX, line 12: This is an ugly workaround 🥲

antfu: I want to work on that to have `codeToHast` able to load grammars on demand - but that would requires quote some changes to the codebase to provide an async version of shiki - which I not yet have time to work on it properly 😇

PR #2045, opened 2025-02-02
docs: explain how to install icon sets

Hofer-Julian:
This is mostly taken from UnoCSS' docs: https://unocss.dev/presets/icons#icon-customizations


KermanX (approved):
Thank you very much~

Issue #2060, opened 2025-02-11
Title is missing from Frontmatter type


c-harding:
**Describe the bug**

The type `Frontmatter` does not contain the field `title`, despite it being present in the docs. 

**Minimal reproduction**

Define the following layout `default.vue`.
```vue
<script setup lang="ts">
import { Frontmatter } from "@slidev/types";
import Frame from "../components/frame.vue";

const { frontmatter } = defineProps<{ frontmatter: Frontmatter }>();
</script>

<template>
  <div>
    {{ frontmatter.title }}
    <slot />
  </div>
</template>
```

Render the slide:

```md
---
title: What is  Slidev?
---

Slidev is a slide maker and presentation tool designed for developers. It includes the following features:

- 📝 **Text-based** - focus on your content with Markdown, then style it later
- 🎨 **Themable** - themes can be shared and reused as npm packages
- 🧑‍💻 **Developer Friendly** - code highlighting, live coding with autocompletion
- 🤹 **Interactive** - embed Vue components to enhance your expressions
- 🎥 **Recording** - built-in recording and camera view
- 📤 **Portable** - export to PDF, PPTX, PNGs, or even a hostable SPA
- 🛠 **Hackable** - virtually anything that's possible on a webpage is possible in Slidev

<br>
<br>

Read more about [Why Slidev?](https://sli.dev/guide/why)
```

This fails to typecheck (`Property 'title' does not exist on type 'Frontmatter'.ts-plugin(2339)`), but runs correctly.

**Docs**: https://sli.dev/custom/#frontmatter

```yaml
# override the title for the <TitleRenderer> and <Toc> components
# only if `title` has also been declared
title: undefined # or string
```

**Environment**

- Slidev version: 51.1.0
- Typescript: ~5.7.3
- OS: Mac

If you are using Slidev globally (i.e. `npx slidev` or `npm i -g slidev`), please try to reproduce the issue in a local project (i.e. `npm create slidev@latest`).

PR #2063, opened 2025-02-13
fix: frontmatter types and vscode schemas

mnenie:
there was an issue where the `title` property was missing in the `Frontmatter` interface, and the `dragPos` property had an incorrect type due to the possibility to use as `Record<string, string>` or `Record<string, string>[]`.

so, this pr resolves these issues and also includes updates to the vscode schemas (which also fix the missing info by adding backticks)
missing info:
Снимок экрана 2025-02-13 170609

additionally, the docs has been updated.

resolves: #2060

### DIFF
--- a/packages/types/src/frontmatter.ts
+++ b/packages/types/src/frontmatter.ts
@@ -1,7 +1,7 @@
 import type { BuiltinLayouts } from './builtin-layouts'
 import type { SlidevThemeConfig } from './types'
 
-export interface Headmatter extends HeadmatterConfig, Frontmatter {
+export interface Headmatter extends HeadmatterConfig, Omit<Frontmatter, 'title'> {
   /**
    * Default frontmatter options applied to all slides
    */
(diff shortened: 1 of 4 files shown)


packages/types/src/frontmatter.ts
```diff
@@ -321,7 +326,7 @@ export interface Frontmatter extends TransitionOptions {
    *
    * See https://sli.dev/features/draggable
    */
-  dragPos?: Record<string, string>[]
+  dragPos?: Record<string, string> | Record<string, string>[]
```
kermanx, line 329: I think this should be just `Record<string, string>`:

```suggestion
  dragPos?: Record<string, string>
```

kermanx (changes requested):
Thank you very much~

packages/types/src/frontmatter.ts
mnenie: @kermanx I think so too, and I use the object myself, but through the array we can also use it 👀

PR #2089, opened 2025-02-25
fix: aspectRatio type in schema

jpoehnelt:
```
Incorrect type. Expected "number".yaml-schema: 
file:///.../antfu.slidev-51.3.0/schema/headmatter.json
Aspect ratio for slides should be like 16/9 or 1:1
```


antfu (approved):

kermanx (changes requested):
It seems that https://github.com/slidevjs/slidev/blob/dd883b792fdd50d7ed342e78b26e2c2307fef4e0/packages/parser/src/config.ts#L95 has normalized the value. All the additions of `parseAspectRatio` in this PR seem to be unnecessary. Correct me if I am wrong

PR #2094, opened 2025-03-02
fix: load raw slide content in slide editor

### LINKED ISSUE #1579: Interaction between preprocessor/preparser and slide editor
**Describe the bug**
This one is a little tricky and I apologize if misunderstanding how things should work.  The preparser extensions (documented [here](https://sli.dev/custom/config-parser)) imply that your .md file is read in and then processed prior to be rendered.  However, when using slide edit mode, the preprocessed file is written back out to disk.  For example, in the [example 1](https://sli.dev/custom/config-parser#use-case-1-compact-syntax-top-level-presentation) where you replace your compact syntax notation, the idea is to make your markdown shorter.  However, if you were to run this, then open the code in the slide editor would immediately overwrite your compact .md file with the expanded versions of these macros.

**To Reproduce**

1. add a `setup/preparser.ts` file with something that adds new lines to the .md during preprocessing.

```
import { definePreparserSetup } from '@slidev/types'

export default definePreparserSetup(({ filepath, headmatter, mode }) => {
  return [
    {
      transformRawLines(lines) {
        // if front matter exists then insert at the top
        // assumes you correctly ended the front matter with '---'
        if (lines[0] === '---') {
          lines.splice(1, 0, `filename: ${filepath}`)
        } else {
          // otherwise add frontmatter to first slide containing the file path
          lines.unshift(`---`)
          lines.unshift(`filename: ${filepath}`)
          lines.unshift(`---`)
        }
      },
    },
  ]
})
```

2. open the slide in the slide editor
3. observe that the .md file now has `filename: XXX` in the front matter defined twice (also there start being errors from the front matter preprocessor because in this case the change introduced two variables with the same name).

My expectation is that opening the slide editor while using the preparser shouldn't interact.  Alternatively, this issue could be expected behavior in which case I'd mention that people should view the preprocessor as being applied multiple times to your file in the event you open the slide editor (so the code needs extra logic to handle that).  Alternatively, the slidev should somehow write out to disk only the changes to the file from the slide editor but none of the changes rendered by the preparser. 

- Slidev version: v0.49.0-beta.6

jgosmann:
Fixes #1579. Otherwise, the editor would load the preprocessed content and overwrite the unprocessed content.

### DIFF
--- a/packages/client/internals/SideEditor.vue
+++ b/packages/client/internals/SideEditor.vue
@@ -26,7 +26,7 @@ watch(
     if (!isInputting.value) {
       note.value = (v?.note || '').trim()
       const frontmatterPart = v?.frontmatterRaw?.trim() ? `---\n${v.frontmatterRaw.trim()}\n---\n\n` : ''
-      content.value = frontmatterPart + (v?.content || '').trim()
+      content.value = frontmatterPart + (v?.source.contentRaw || '').trim()
       dirty.value = false
     }
   },
--- a/packages/parser/src/core.ts
+++ b/packages/parser/src/core.ts
@@ -110,6 +110,7 @@ export function parseSlide(raw: string, options: SlidevParserOptions = {}): Omit
     level,
     revision,
     content,
+    contentRaw: content,
     frontmatter,
     frontmatterStyle: matterResult.type,
     frontmatterDoc: matterResult.doc,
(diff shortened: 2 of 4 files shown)


jgosmann:
The failing tests seem unrelated to me. They also fail for me when running them locally from the `main` branch.

jgosmann:
Will this be merged? Or would you like to see issue #1579 solved in some other way?

(The two failed checks seem to have timed out for reasons unrelated to the PR.)

kermanx (approved):

antfu (approved):

Issue #1932, opened 2024-11-03
Support Open Graph protocol


kaakaa:
**Is your feature request related to a problem? Please describe.**

It would be nice if the generated slides has `og` properties of [Open Graph protocol](https://ogp.me/), making them easier to share on blog, SNS, etc.

**Describe the solution you'd like**

Like `download` config, setting `ogp` config enabled will inject `og` properties to `index.html` when building a slide. `ogp` config would be needed to have the URL to deploy to, because some `og` properties require absolute URL and the URL isn't determined when building.

e.g.:

If `ogp` setting is enabled,

```yaml
---
ogp:
  url: https://slides.exaple.com/
  image: https://slides.exaple.com/ogp.png
---
```

`og` properties will be injected to `index.html` in [setupIndexHtml](https://github.com/slidevjs/slidev/blob/3be1191199fc2d0814f60aed139c81a299477a79/packages/slidev/node/setups/indexHtml.ts#L22), and  og-image can be generated in [build.ts](https://github.com/slidevjs/slidev/blob/3be1191199fc2d0814f60aed139c81a299477a79/packages/slidev/node/commands/build.ts#L78) with `{format: 'png', range: 1}` options.

```ts
const ogImage = ogp.image ? ogp.image : `${ogp.url}/assets/ogp/image.png`
...
head += [
  ogp && `<meta property="og:title" content="${getSlideTitle(data)}" />`,
  ogp && `<meta property="og:type" content="video.movie" />`,
  ogp && `<meta property="og:url" content="${ogp.url}" />`,
  ogp && `<meta property="og:image" content="${ogImage}" />`,
]
...
```

(However, my concern is that this way fixes the deployment location at build time, so it would be good if URLs such as og:url could be dynamically determined according to the deployment location...)

**Describe alternatives you've considered**

Now, I'm running [a script](https://github.com/kaakaa/slidev-resources/blob/main/scripts/setup-spa.ts) on GitHub Actions to generate `index.html` with `og` properties. It works fine, but I would love for Slidev to have this function;

antfu:
I'd be happy to have this feature. I think it would be better to have `og` instead of `ogp` tho. For the URL I am not sure as Slidev generates SPA without server, where we don't know where the page is deployed, I feel it's better not to do any special handling and leave them to the users.

As tagged, PR is welcome if anyone wants to work on it

PR #2096, opened 2025-03-03
feat: support open graph protocol

leochiu-a:
# Description

Support Open Graph Protocol in the headmatter. Only adding support for the following attributes:

```
seoMeta:
  ogTitle: Slidev Starter Template
  ogDescription: Presentation slides for developers
  ogImage: https://cover.sli.dev
  ogUrl: https://example.com
  twitterCard: summary_large_image
  twitterTitle: Slidev Starter Template
  twitterDescription: Presentation slides for developers
  twitterImage: https://cover.sli.dev
  twitterSite: username
  twitterUrl: https://example.com
```

# Test Plans

- The SEO meta tag should be visible in the view source page.

# 

# Related issue

- https://github.com/slidevjs/slidev/issues/1932

### DIFF
--- a/demo/starter/slides.md
+++ b/demo/starter/slides.md
@@ -21,11 +21,8 @@ transition: slide-left
 # enable MDC Syntax: https://sli.dev/features/mdc
 mdc: true
 # open graph
-seoMeta:
-  ogTitle: Slidev Starter Templat
-  ogDescription: Presentation slides for developers
-  ogImage: https://cover.sli.dev
-  ogUrl: https://example.com
+# seoMeta:
+#  ogImage: https://cover.sli.dev
 ---
 
 # Welcome to Slidev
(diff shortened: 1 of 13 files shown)


antfu:
I wonder if we should use something more comprehensive like https://unhead.unjs.io to support more tags other than only og

leochiu-a:
@antfu Do you think we need to implement all Open Graph attributes using useSeoMeta, or should we provide only the essential tags, such as the following? 

- og:title
- og:desription
- og:image
- og:url
- twitter:card
- twitter:site
- twitter:title
- twitter:description
- twitter:image
- twitter:url

Then, web can provide a `seoMeta` attribute on headmatter which extends the `MetaFlat` type from `unhead`

antfu:
Yeah that looks great, thanks!

ogadra (approved):
Thanks for developing this! I would love to have this feature as soon as possible!

It seems difficult to implement without using `transformHtmlTemplate`, so I’m okay with using an unstable version of `unhead` if necessary.

leochiu-a:
Hi, @antfu. Could you take a look when you have time?

PR #2194, opened 2025-06-03
fix: note style

### LINKED ISSUE #1712: Improve monospace and list formatting of notes
**Describe the bug**

### Monospace
Backticks of monospace notes are shown both in presenter mode and in exported notes:
Screenshot 2024-06-29 at 00 14 30
Screenshot 2024-06-28 at 23 43 18

I think that these monospace texts should look the same as in slides:
Screenshot 2024-06-28 at 23 43 38

### Lists
I think that the spacing after indented note list items should be the same as after non-indented ones. (See the first two illustrations above.) I suggest the same for spacings _before_ indented list items:
Screenshot 2024-06-29 at 00 18 59

**Minimal reproduction**

1. Export the notes in my presentation:
    ```bash
    git clone https://github.com/gy-mate/kalauz
    cd kalauz/presentation
    git checkout 299e314a
    npx slidev
    ```
4. Open http://localhost:3030/presenter/16
5. See the monospace text in point no. 4 and the spacing after it
6. Open http://localhost:3030/presenter/18?clicks=2
7. See the spacing before the list element

**Environment**

- Slidev version: `0.49.12` (global)
- Browser: Safari `17.5`
- OS: macOS `14.5`

### LINKED ISSUE #2145: Broken note rendering for nested lists
**Describe the bug**

The rendering of nested lists in notes is broken, also if the elements have links or not.
The nested lists should have the same spacing between each element.

Image

**Minimal reproduction**

Add the following note to a slide

```markdown
<!--
Nested list with links:

- [Content](https://github.com/slidevjs/slidev)
- [Content](https://github.com/slidevjs/slidev)
  - [Content](https://github.com/slidevjs/slidev)
- [Content](https://github.com/slidevjs/slidev)

Nested list without links:

- Content
- Content
  - Content
- Content
-->
```

**Environment**

- Slidev version: 51.5.0
- Browser: Firefox/Chrome
- OS: Windows 11

yuyinws:
resolve #2145
resolve #1712

## Description
This PR fix `lists` and `code block` render styles on presenter notes.

### Before

before

### After

after

### DIFF
--- a/packages/client/uno.config.ts
+++ b/packages/client/uno.config.ts
@@ -53,7 +53,14 @@ export default defineConfig({
   presets: [
     presetWind3(),
     presetAttributify(),
-    presetTypography(),
+    presetTypography({
+      cssExtend: {
+        ul: {
+          'margin': '0',
+          'line-height': 1.75,
+        },
+      },
+    }),
     /* Preset Icons is added in ../node/setups/unocss.ts */
   ],
   transformers: [
--- a/packages/client/styles/code.css
+++ b/packages/client/styles/code.css
@@ -123,11 +123,17 @@ html:not(.dark) .shiki span {
 }
 
 /* Inline Code */
+.slidev-note :not(pre) > code,
 .slidev-layout :not(pre) > code {
   font-size: 0.9em;
   background: var(--slidev-code-background);
   border-radius: var(--slidev-code-radius);
   --uno: font-light py-0.5 px-1.5;
+}
+
+.slidev-note :not(pre) > code:after,
+.slidev-note :not(pre) > code:before {
+  content: '';
 }
 
 .slidev-layout :not(pre) > code:before {


packages/client/uno.config.ts
```diff
@@ -53,7 +53,14 @@ export default defineConfig({
   presets: [
     presetWind3(),
     presetAttributify(),
-    presetTypography(),
+    presetTypography({
+      cssExtend: {
```
antfu, line 57: Why do we need this tho?

yuyinws: Because the `note block` use unocss's typography style. Cause nested lists have different spacing:

image

yuyinws: Maybe we should remove the `line-height` overwrite?

antfu: I am mostly worried about if this change would break user's style

yuyinws: Oh I am not considering this situation before, overwrite it use `slidev-note` selector now, it should be safe for user.

kermanx (approved):

yuyinws:
Hi @antfu, sorry for the pin, want to know what block this PR being merged? I'm happy to help.

PR #2208, opened 2025-06-22
fix: add optional chain to prevent nullish broken

### LINKED ISSUE #2047: Inconsistent Router Injection Error with Additional Resolvers
**Describe the bug**

When using additional component resolvers, the application becomes blank after a while, with the following browser console errors:

```
[Vue warn]: injection "Symbol(router)" not found.
[Vue warn]: injection "Symbol(route location)" not found.
Uncaught (in promise) TypeError: Cannot read properties of undefined (reading 'currentRoute')
```

This is very weird problem. Sometimes the slides work fine, everything including the PrimeVue components are displayed correctly, but after a few seconds app div will become blank. Sometimes the slides don't load at all.

**Minimal reproduction**

Note that this problem is not reproduced consistently. Here's what my customization looks like:

```ts
// main.ts
import { defineAppSetup } from "@slidev/types";
import PrimeVue from "primevue/config";
import Aura from "@primevue/themes/aura";

export default defineAppSetup(({ app }) => {
  // TODO: support specifying themes in head matter
  app.use(PrimeVue, {
    theme: {
      preset: Aura,
      options: {
        // PrimeVue uses system dark mode selector by default, which conflicts
        // with the toggle in Slidev
        darkModeSelector: ".dark"
      }
    },
  });
});
```

```ts
/// <reference types="@slidev/types" />

import { defineConfig } from "vite";
import { PrimeVueResolver } from "@primevue/auto-import-resolver";

export default defineConfig({
  slidev: {
    components: {
      resolvers: [PrimeVueResolver()],
    },
  },
});
```

1. Install the addon using `npm install slidev-addon-prime`
2. Add the addon to `slides.md`:

```yaml
---
addons:
  - prime
---
```

3. (optional) add some PrimeVue components

```html
<div class="card flex justify-center flex-wrap gap-4">
  <Button label="Primary" />
  <Button label="Secondary" severity="secondary" />
</div>
```

4. Run `slidev`.

**Potential Causes**

Interference between the router injection and other resolvers.

ShenQingchuan:
I have encountered this issue in my own PPT project (which can be a reproduction repo): https://github.com/ShenQingchuan/vue-vine-ppt-2025

image

But it'll be OK after restarting the slidev dev server.

I think it might be caused by my Vue Vine plugin using in setup/main.ts, maybe it's applied before Vue Router and the router instance was not initialized at that moment? I'm not quite sure ...

But I've tried and approved that this line of change works as expected, which is adding optional chain operators.

fix #2047

### DIFF
--- a/packages/client/composables/useNav.ts
+++ b/packages/client/composables/useNav.ts
@@ -277,7 +277,7 @@ const useNavState = createSharedComposable((): SlidevContextNavState => {
 
   const query = computed(() => {
     // eslint-disable-next-line ts/no-unused-expressions
-    router.currentRoute.value.query
+    router?.currentRoute?.value?.query
     return new URLSearchParams(location.search)
   })
   const isPrintMode = computed(() => query.value.has('print') || currentRoute.name === 'export')


kermanx (approved):
Seems to be https://github.com/slidevjs/slidev/issues/2047

kermanx (approved):
Seems to be https://github.com/slidevjs/slidev/issues/2047

PR #2226, opened 2025-07-13
feat: auto gen og

naitokosuke:
# Implement automatic og-image generation from first slide

Hi Slidev maintainers.

Thank you so much for your amazing work on Slidev.
I absolutely love this project!

This is my first feature implementation PR for Slidev, so please let me know if there's anything that needs improvement or if I'm causing any trouble.

I'm happy to make any necessary changes.

## Motivation

Currently, Slidev users need to manually create og-images for social media sharing.
The existing workflow requires:

- Export first slide as PNG
  - Run `slidev export --format png --range 1` to generate first slide image

- Rename and move file
  - Rename `01.png` to `og-image.png` and place in appropriate location
- Host image somewhere
  - Upload to CDN, GitHub, or image hosting service

- Configure URL in frontmatter
  ```yaml
  seoMeta:
    ogImage: https://example.com/og-image.png
  ```

## Usage

```yaml
---
# slides.md frontmatter
seoMeta:
  ogImage: auto # Automatically generates og-image from first slide
---
```

### DIFF
--- a/.github/workflows/smoke.yml
+++ b/.github/workflows/smoke.yml
@@ -107,6 +107,10 @@ jobs:
         working-directory: ../temp/slidev-project
         if: ${{ matrix.pm != 'yarn' }}
 
+      - name: Install Playwright browsers
+        run: pnpx playwright install chromium
+        working-directory: ../temp/slidev-project
+
       - name: Install project (yarn)
         run: yarn add /tmp/slidev-pkgs/cli.tgz playwright-chromium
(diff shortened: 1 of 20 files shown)


antfu:
Thanks! That's a great idea!

I think we could improve it even more:

- We should by default support using `og-image.png` on the project root
- With this auto-gen feature, we generate `./og-image.png` on the project if it doesn't exist, so users could also commit it to the source control or make modifications to avoid repetitive exporting on CI

PR #2263, opened 2025-09-01
fix: display IME characters during composition

### LINKED ISSUE #2262: bug: characters not displayed during IME composition
**Describe the bug**

First, thank you for the awesome tool! I use Slidev for my regular presentations.

Slidev has an [integrated editor feature](https://sli.dev/features/side-editor) to edit the slide content in the sidebar. However, when using the Japanese IME, the characters during convertion do not appear on the screen and only update after conversion is complete.

https://github.com/user-attachments/assets/014381e1-63f4-42da-a62d-eabbb7c28cd0

Currently, Slidev use `keydown` event to capture characters, but should use `compositionupdate`. I will submit a PR to resolve this issue.

https://github.com/slidevjs/slidev/blob/5637013f7a6230e9de921185a7381c9cff3f4f8c/packages/client/internals/SideEditor.vue#L56-L61

**Minimal reproduction**

1. `pnpm create slidev`
2. `pnpm dev`
3. Turn on Japanese IME and put characters, but they do not appear.

**Environment**

- Slidev version: v52.1.0
- Browser: Microsoft Edge
- OS: Windows

ras0q:
resolve #2262

Fixed it so that during IME input, values are updated only within the Editor, and once the IME confirms the input, the value is passed to the caller.

|before|after|
|-|-|
image|image

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,27 +10,30 @@ jobs:
     permissions:
       id-token: write
       contents: write
+
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-      - uses: actions/setup-node@v4
+
+      - uses: pnpm/action-setup@v4
+
+      - uses: actions/setup-node@v5
         with:
           node-version: lts/*
           registry-url: https://registry.npmjs.org/
+
       - run: npx changelogithub
         env:
           GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
-      - run: npm i -g @antfu/ni
-      - run: nci
-      - run: nr ci:publish
-        env:
-          NODE_AUTH_TOKEN: ${{secrets.NPM_TOKEN}}
-          NPM_CONFIG_PROVENANCE: true
+
+      - run: npm i -g npm@latest
+      - run: pnpm i
+      - run: pnpm run ci:publish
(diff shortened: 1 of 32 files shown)


antfu:
I am afraid this is too complex and would be hard to understand and maintain. Don't we have a better way to do that? Or at least I think it would be better to move them into a headless component

ras0q:
I didn't know `InputEvent` had an `isComposing` property.
https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/isComposing

Furthermore, since the `target` property of `InputEvent` contains the character being input via IME, we could handle most processing in `onInput`.
(However, since `onInput` doesn't fire when IME input ends, we need to use `onCompositionEnd`...)

How does this look? Should we further separate this into a headless component?

```typescript
const localContent = ref(content.value)
watch(content, (v) => {
  if (v !== localContent.value) {
    localContent.value = v
  }
})

function onInput(e: Event) {
  if (!(e instanceof InputEvent) || !(e.target instanceof HTMLTextAreaElement)) {
    return
  }

  if (e.isComposing) {
    localContent.value = e.target.value
  }
  else {
    content.value = e.target.value
  }
}

function onCompositionEnd() {
  content.value = localContent.value
}
```

antfu:
Yeah let's do a headless component. Thanks!

ras0q:
@antfu I fixed it. Could you please review it again?

antfu (approved):

PR #2273, opened 2025-09-10
Produce tagged PDFs

CrossR:
Opening this more for discussion, as I'm sure there is plenty of things I'm not considering...so didn't want to spend a bunch of time on it then realise I'd missed a bunch.

Chrome a while ago added the ability to export tagged PDFs, then a bit later, puppeteer added the functionality to use that as well: https://github.com/puppeteer/puppeteer/issues/7509.
That has both the feature in chrome linked, then later the PR to add it to puppeteer.

Enabling it makes the produced PDF documents more accessible, and means things like `"alt"` text on images are exported:

Before change:

Screenshot 2025-09-10 at 13 58 12

After Change:

image

With the code in the demo slides just being:

```md
<img border="rounded" src="https://github.com/slidevjs/themes/blob/main/screenshots/theme-default/01.png?raw=true" alt="Image of an alternative theme,  with a water background.">
```

This is obviously not a complete PR, but does show that with an extra option passed over in 3 more places, the produced PDFs (if some time is spent on adding alt text to ones talks too...) could be a lot more usable via screen readers and such.

There is probably a question of "Are the PDFs even readable enough for this to matter" which I can't really answer, but it would be cool to see this change added.

Things I don't know:

 - Browser and browser version support?
 - Enabled by default or not? Option for it, or just always on etc.
 - Anything else I'm missing? Can it be enabled via Chrome /export or only the CLI etc.

As I say...this is mostly for a discussion, so if someone with more experience / time already has answers to my questions above, and wants to work on it instead, that is fine with me! I'd just like my slides to be slightly more readable for my colleagues.

Apologies if I missed an open issue / PR on this....I searched a few terms but didn't find anything.

### DIFF
--- a/packages/slidev/node/commands/export.ts
+++ b/packages/slidev/node/commands/export.ts
@@ -154,6 +154,7 @@ export async function exportNotes({
     },
     printBackground: true,
     preferCSSPageSize: true,
+    tagged: true,
   })
 
   progress.stop()
@@ -358,6 +359,7 @@ export async function exportSlides({
         pageRanges: '1',
         printBackground: true,
         preferCSSPageSize: true,
+        tagged: true,
       })
       buffers.push(pdf)
     }
@@ -400,6 +402,7 @@ export async function exportSlides({
       },
       printBackground: true,
       preferCSSPageSize: true,
+      tagged: true,
     })
 
     // Edit generated PDF: add metadata and (optionally) TOC


jrg94:
I was able to verify that this change works on a much older version of slidev (`0.37.1`). I have not updated because I use the Unicorn theme, and it has not been modernized for newer versions of slidev. In any case, manually editing in the `tagged: true` lines worked for me. 

As an educator, the tagged PDF is a lifesaver. We're required now to meet higher accessibility standards with our materials, and this change was able to get me a better accessibility score on my existing slide decks. Normally, I would see something like this:
image

This pull request gets me much closer to a "passing" score (though, I'm not sure decorative images like the logo in this case should even have alt tags):
image

ani-per:
@antfu @kermanx Just bumping this a bit as it's a very tiny PR with a significant upgrade in value, so could be reviewed pretty quickly.

c-harding:
It looks like this flag was originally added to Playwright because of the increase in file size. How much does this update affect your PDF file size? Do you think it would be worth adding a new flag to the export commands to enable/disable this feature?

https://github.com/microsoft/playwright/issues/13917

PR #2292, opened 2025-09-27
fix: prevent Twoslash poppers from persisting between slides

### LINKED ISSUE #2202: Slidev twoslash popover keep being open between slides
**Describe the bug**

Slidev twoslash popover keep being open between slides

**Minimal reproduction**

```

/```
const a = 5;
//       ^?
/```

---

# Other slide 

---

```

Image

Image

1. Go to  a slide with a twoslash `//  ?^` to show the type information. A popover opens with the type information.
2. Click the right arrow to go to the next slide
3. See the same popover not closing

You can use https://sli.dev/new to create a new project to reproduce the issue.

**Environment**

- Slidev version: 51.2
- Browser: Chrome
- OS: MacOS

If you are using Slidev globally (i.e. `npx slidev` or `npm i -g slidev`), please try to reproduce the issue in a local project (i.e. `npm create slidev@latest`).

elecmonkey:
Prevent Twoslash poppers from persisting when navigating across slides by watching `currentSlideRoute`.

Fixes #2202


elecmonkey:
The changes make a test fail. I think I should not fix this issue by using v-if, as it will destroy the components when navigation happens.

elecmonkey:
The fixed popers displayed by `^?` in Twoslash doesn’t work well in Slidev. When adjacent slides are preloaded—or when all slides are preloaded after three seconds— `^?` poppers appear, even on slides they don’t belong to. Moreover, once these poppers show up, navigate to other slide pages doesn’t close them unless they were manually closed.

However, It isn't worthwhile to disrupt the current rendering or preload flow just to fix this. Is there a better solution under these constraints?
As for the issue #2202 (which could affect the presentation process), I believe the changes in `useNav.ts` can resolve that. But for the poppers showing during preload (which results in the slide being cluttered with popups on first load), my current thought is to close them after poping-up with a setTimeout. That said, this problem is generally less critical during a presentation compared to the one discussed in the issue #2202 , since it only happens when slides are initially loaded.

My thoughts might be a bit scattered — want to hear what the community and maintainers think.

packages/client/internals/SlidesShow.vue
```diff
@@ -39,12 +39,18 @@ watchEffect(() => {
   preloadRoute(currentSlideRoute.value)
   preloadRoute(prevRoute.value)
   preloadRoute(nextRoute.value)
+  setTimeout(() => {
+    hideAllPoppers()
+  }, 150)
```
antfu, line 44: This timeout time is too arbitrary. Could we find a better way to do that instead of relying on timeout?\

elecmonkey: It stems from FloatingVue's behavior. when v-popper directives are rendered to the DOM, FloatingVue immediately creates popper instances and shows them based on the `:shown='true'`...   so I think maybe we could implement a custom Shiki transformer to intercept the Twoslash rendering process, and try to transform the code

elecmonkey:
Seems like not the cleanest solution, but I think it works beautifully now!

https://github.com/user-attachments/assets/3f99f218-e2c3-487c-92fd-c9defe2f62a7

packages/slidev/node/syntax/transform/twoslash-conditional.ts
```diff
@@ -0,0 +1,17 @@
+import type { ShikiTransformer } from 'shiki'
+
+/**
+ * Custom Twoslash transformer that replaces :shown='true' with conditional logic
+ * to prevent poppers from being shown during preload
+ */
+export function transformerTwoslashConditional(): ShikiTransformer {
```
antfu: We should use the build-time flag rather than a custom codemod (which is unstable), like:

https://github.com/slidevjs/slidev/blob/6750606dd1b7446ccb8de5030f913d3141f1e451/packages/slidev/node/options.ts#L119-L130

antfu (changes requested):

dahlia:
Hi @elecmonkey and @antfu,

I've been experiencing the issue this PR fixes (#2202), and I'd love to see this merged! The Twoslash popover persisting between slides is quite noticeable during presentations.

Just wanted to check in on the status—is there anything that needs to be addressed following @antfu's feedback on the build-time flag approach? Happy to help with testing or anything else if needed.

Thanks for working on this!

kermanx (approved):

PR #2295, opened 2025-09-30
[DRAFT] Add speedrunning-inspired split timing support

cmgriffing-at-gitkraken:
## Motivation

When I give talks, I find it hard to know how I am doing for pacing.
The timer helps, but knowing how far off I am at specific points in the presentation would be very helpful.

In the speedrunning community, they use tools that look a bit like this:
image

## Demo code

This PR is specifically for getting the idea and feature discussed. The implementation is very naively done and I will happily write it in a more idiomatic way rather than the event listener approach.

## Questions

1. Is this a feature that would be accepted (assuming it is implemented better)?
2. Do you have any specific opinions about where in the presenter view this should live?
3. What would be the ideal implementation? A custom `useSplits` hook? This would be useful for a custom component that users may want to put on their slides directly if they don't use presenter view.

### DIFF
--- a/demo/starter/slides.md
+++ b/demo/starter/slides.md
@@ -45,6 +45,7 @@ The last comment block of each slide will be treated as slide notes. It will be 
 
 ---
 transition: fade-out
+split: 30
 ---
 
 # What is Slidev?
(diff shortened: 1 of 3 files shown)


antfu:
Would love to see it happend!

antfu:
Working on #2309

PR #2302, opened 2025-10-08
fix: prevent hotkeys when typing in monaco editor

### LINKED ISSUE #2297: Hotkeys triggered when editing monaco editor in Chromium-based browsers
**Describe the bug**

While editing a monaco editor snippet, adding a space would move to the next slide, adding a `f` would enter or exit full screen, adding an `o` opens the overview, etc. This makes it virtually impossible to edit monaco snippets from slides...

This only happens in Chromium based browsers. It works as expected in Firefox and Safari.

**Minimal reproduction**

1. Create a new project

  ```sh
  pnpm create slidev
  cd slidev
  pnpm install
  pnpm approve-builds
  pnpm run dev
  ```
2. Open in a Chromium-based browser, go to slide 15 and type `o` in one of the snippets

**Environment**

- Slidev version: 52.2.4 
- Browser: Chromium 143.0.7449.0 / Arc 1.114.1
- OS: Mac OS 26.0.1

QuentinRoy:
Prevent Slidev hotkeys from triggering when typing in the editor in Chromium-based browsers.

Fixes #2297.

It is unclear whether #2297 originates from Slidev, Monaco, or Chromium. That said, this change simply prevents keydown events originating from the Monaco editor from propagating. I don’t see any cases where allowing this propagation would be desirable.

### DIFF
--- a/packages/client/setup/monaco.ts
+++ b/packages/client/setup/monaco.ts
@@ -4,8 +4,8 @@ import setups from '#slidev/setups/monaco'
 import { createSingletonPromise } from '@antfu/utils'
 import { setupTypeAcquisition } from '@typescript/ata'
 import * as monaco from 'monaco-editor'
-
 import EditorWorker from 'monaco-editor/esm/vs/editor/editor.worker?worker'
+
 // @ts-expect-error missing types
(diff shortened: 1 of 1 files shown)


QuentinRoy:
This does not work, it prevents monaco own shortcuts too like `cmd + z`.

PR #2309, opened 2025-10-20
feat: support setting duration of timer and provide countdown mode

antfu:
Idea from #2295


cmgriffing-at-gitkraken:
I'm glad you liked the idea. I like your solution more too. Thanks

antfu:
Postponed the timesplit feature, this PR is doing some enhancements to the timer and the ability to set overall duration

PR #2343, opened 2025-11-12
chore: bump pnpm v10.21 + enable trust policy

iiio2:
pnpm 10.21 has a very good security feature for enabling trust policy. maybe we can use it here. 
https://pnpm.io/settings#trustpolicy


iiio2:
Thanks @antfu .

PR #2359, opened 2025-11-15
fix(vscode): should throttle project rescan

### LINKED ISSUE #2213: vscode slidev extension exhausts resources (memory, cpu) when running npm install
**Describe the bug**

When running `npm install` in one of my workspaces with a large number of dependencies, my memory quickly gets maxed out and my only solution is to kill vscode. I traced the issue to the slidev extension, which causes _many_ rg (ripgrep) processes to be launched. Disabling the extension makes `npm install` work again.

https://github.com/microsoft/vscode/wiki/Search-Issues#slow-search-rg-running-for-a-long-time-or-consuming-lots-of-cpumemory

### LINKED ISSUE #2290: vscode extensions: Creating a large amount of markdown files at once exhausts cpu
When using a tool that creates a large amount of markdown files at once. It spawns an insane amount of ripgrep processes which ultimately freezes the computer.

**Minimal reproduction**

Use something that creates a lot of markdown files.
For example: typedoc-plugin-markdown plugin for typedoc.
Run the typedoc command and observe the rg processes being created.

**Environment**

- Slidev version: latest
- Browser: N/A
- OS: macos

kermanx:
fix #2213, fix #2290

It works, but can't be released before #2207 is merged.

### DIFF
--- a/packages/vscode/src/projects.ts
+++ b/packages/vscode/src/projects.ts
@@ -1,4 +1,5 @@
 import type { LoadedSlidevData } from '@slidev/parser/fs'
+import type { Uri } from 'vscode'
 import { existsSync } from 'node:fs'
 import { basename, dirname } from 'node:path'
 import { slash } from '@antfu/utils'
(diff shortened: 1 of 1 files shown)


kermanx:
In fact, I now think another rewrite of the extension is required 😓

In the process of developing https://github.com/kermanx/p2p-live-share, I realized that I made a few mistakes in file updating in this extension.

antfu:
@kermanx If you want to do the rewrite, I am definitely up to that! And maybe let's have it as a seperate repo in the org? WDYT?

Issue #2337, opened 2025-11-07
Unindented list inside HTML tag causes error


RyotaUshio:
I'm not sure if I should file this issue to this repository or [slidevjs/prettier-plugin](https://github.com/slidevjs/prettier-plugin). Excuse me if this is the wrong repo.

**Describe the bug**

The following syntax is proper. (Notice that the inner list is indented)

```
<div>
  <div>

  - A
  - B

  </div>
</div>
```

If I remove the indent (**which `prettier-plugin-slidev` does**), I got an error:

```
<div>
  <div>

- A
- B

  </div>
</div>
```

Error:

```
[plugin:vite:vue] Element is missing end tag.
/Users/ryotaushio/code/slidev-repro/slides.md__slidev_3.md:11:1
9  |  <p>A</p>
10 |  </li>
11 |  <li>
   |   ^
12 |  <p>B</p>
13 |  </div>
    at createCompilerError (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:1364:17)
    at emitError (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:2991:5)
    at Object.onclosetag (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:2350:13)
    at Tokenizer.stateInClosingTagName (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:785:16)
    at Tokenizer.parse (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:1143:16)
    at Object.baseParse (/Users/ryotaushio/code/slidev-repro/node_modules/.pnpm/@vue+compiler-core@3.5.23/node_modules/@vue/compiler-core/dist/compiler-core.cjs.js:3030:13)
[...]

PR #2371, opened 2025-11-18
Unindented list inside HTML tag causes error 

Bitshifter-9:
Added transformHtmlListIndent to preserve list indentation inside raw HTML blocks in slide Markdown, preventing Vue compile errors caused by Prettier removing spaces.

Integrated the transformer early in the Markdown pipeline.

Added regression tests and a snapshot for the

+ list scenario to ensure future safety.
this closes issue #2337 
@kermanx


kermanx:
Thank you for working on this. The code is fine, but I am not sure if we are in the right direction, as I said in https://github.com/slidevjs/slidev/issues/2337#issuecomment-3544850837

PR #2379, opened 2025-11-24
fix: force end layout to use dark theme

elecmonkey:
This PR updates the built-in end layout to always render in dark mode.

Previously, the end layout only applied a black background via its own styles, but the surrounding theme context could still be in light mode. As a result, components rendered inside the end layout (such as the built-in PoweredBySlidev component using the text-main class) would pick up light-theme colors, which are not visually correct on a black background.


kermanx (approved):

antfu:
I think it might be better that we don't force the black background nor the dark mode? I think it might be a case where people want light mode end layout?

kermanx:
> I think it might be better that we don't force the black background nor the dark mode? I think it might be a case where people want light mode end layout?

I think we should instead update the default style for `.slidev-layout.end` with `--uno: dark`.

PR #2389, opened 2025-11-30
fix: use slash utility for relative globs

kermanx:


antfu:
Maybe we should switch to `pathe` later

PR #2400, opened 2025-12-09
chore: add resolver test, configue vitest projects

AlbertBrand:
Add a proper test for the `createResolver` logic, as I want to automatically prevent the regression. 

Also, when running the test locally via VSCode the `vitest` extension did some incorrect assumption which tests to run, which is fixed by explicitly stating the testable projects.

### DIFF
--- a/packages/slidev/node/resolver.test.ts
+++ b/packages/slidev/node/resolver.test.ts
@@ -1,7 +1,7 @@
 import { beforeEach, describe, expect, it, vi } from 'vitest'
 import { createResolver, getRoots } from './resolver'
 
-const mocks = vi.hoisted(() => {
+const vitefuMocks = vi.hoisted(() => {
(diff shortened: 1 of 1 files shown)


AlbertBrand:
@antfu @kermanx This commit with upgrades https://github.com/slidevjs/slidev/commit/d11ddbdb361b2cffb534666c4fa3fd08ae4b2182 seems to have broken builds. When I do a `pnpm i` (with or without force) with that commit checked out and then run `pnpm build` I get:

```
> @52.10.1 build /Users/albertbrand/dev/slidev
> pnpm -r --filter=./packages/** --parallel run build

No matches found: "--filter=./packages/**"
 ELIFECYCLE  Command failed with exit code 1.
```

Maybe from the pnpm upgrade?

antfu (approved):

PR #2403, opened 2025-12-10
docs: fix docs and demo build scripts

AlbertBrand:
Fixes deployment issues (I hope)

### DIFF
--- a/packages/client/composables/useTimer.ts
+++ b/packages/client/composables/useTimer.ts
@@ -16,6 +16,8 @@ export function useTimer() {
     interval.counter.value
     if (state.value.status === 'stopped' || !state.value.startedAt)
       return 0
+    if (state.value.status === 'paused')
+      return state.value.pausedAt - state.value.startedAt
     return Date.now() - state.value.startedAt
   })
   const passed = computed(() => passedMs.value / 1000)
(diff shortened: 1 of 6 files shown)


AlbertBrand:
@antfu @kermanx strange, this change works locally. Is netlify reusing runners? It might be some leftover files issue.

toreis-up:
I could build it locally, too.
It might be Netlify build cache🤔

> If a build fails, it’s worth retrying with a cleared build cache to check if this works better.

https://docs.netlify.com/build/configure-builds/troubleshooting-tips/#build-cache

AlbertBrand:
I have no access to the Netlify setup to refresh the build cache. Maybe @kermanx?

kermanx:
> I have no access to the Netlify setup to refresh the build cache. Maybe @kermanx?

My Netlify account was suspended for no reason :crying_cat_face:

AlbertBrand:
> > I have no access to the Netlify setup to refresh the build cache. Maybe @kermanx?
> 
> My Netlify account was suspended for no reason 😿

That's too bad. Can we ask @antfu for help here? All builds will fail on deployment of the docs until the Netlify cache is cleared, and there's AFAIK no way to trigger cache clear from the pipeline.

Issue #2421, opened 2026-01-04
Revert timer to previous version


Antonytm:
**Is your feature request related to a problem? Please describe.**
The timer became unusable with milliseconds. It takes presenter's attention. That is bad UX. You don't need milliseconds as a presenter. And you get something in the corner that distracts your attention!

**Describe the solution you'd like**
Revert timer to show only seconds. The previous version was fine!

**Describe alternatives you've considered**
Use older version

Antonytm:
```
.text-base.op50 {
  display: none !important;
}
```

for current version...

antfu:
This is open source, and we are welcoming different opinions and different needs. I understand the change might frustrate you, but I also want to bring to your attention that you might want to be aware about your tone when expressing your opinions. 

In general, I am up to add an option or button to toggle the bahaviour, and even better if you want to help thinking about the design/ux and even work on the implementation to send a PR.

Antonytm:
Hi @antfu 

**Sorry for the tone**, I didn't mean to offend you. I meant to highlight that it really may break the presenter's experience and is not only a nitpick. Probably I overdid it here, sorry.

The movement is distracting your attention. I am working on a presentation. I tell a story, and when I forget something, I move my eyes to the second screen with the presenter's notes to get a hint at what is next. And the first thing that catches attention is the quickly moving part. Timer with milliseconds. And it causes the pauses in the presentation and makes the presentation worse...

I can't think of any scenario when you need milliseconds....
That is why I think this feature is useless.

Antonytm:
PR: https://github.com/slidevjs/slidev/pull/2424

It only removes the milliseconds.

I thought about the toggle to hide/show milliseconds. It will overcomplicate settings with the configuration of a feature that is not useful. It is my opinion. If you have a vision of how it should be done, please describe. I will be able to help with the implementation.

PR #2424, opened 2026-01-06
feat: remove milliseconds from timer

Antonytm:


Antonytm:
https://github.com/slidevjs/slidev/issues/2421

PR #2430, opened 2026-01-12
ci: preview release add pm option

btea:
Different developers may use different package management tools. To facilitate copying and execution, commands that satisfy various management tools are generated.

### DIFF
--- a/.github/workflows/cr.yml
+++ b/.github/workflows/cr.yml
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   cr:
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'slidevjs/slidev' }}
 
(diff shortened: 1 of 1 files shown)


.github/workflows/cr.yml
```diff
@@ -5,6 +5,7 @@ on: [push, pull_request]
 jobs:
   cr:
     runs-on: ubuntu-latest
+    if: ${{ github.repository == 'slidevjs/slidev' }}
```
btea: To avoid errors when synchronizing code by forking a repository.

antfu:
I never heard such practice. The changes of forks would be responsible for their own changes, I don't know why this needs to go to the upstream repo. For this repo I am not interested in using other package managers.

PR #2456, opened 2026-02-08
fix: load preparser.ts from theme roots

### LINKED ISSUE #2402: Preparser.ts not loaded from theme
**Describe the bug**

If I create a `setup/preparser.ts` in a theme, it gets ignored.
This might be related to https://github.com/slidevjs/slidev/issues/926

**Minimal reproduction**

1. Create a new slidev project in `apps/slides` via `pnpm create slidev`
2. Create a new theme in `packages/slidev-theme-mytheme` via `pnpm create slidev-theme`
3. In the theme, create a `setup/preparser.ts` besides the `setup/shiki.ts`
4. Add `console.log` to both setup files.
5. The `console.log` from the shiki file gets output, the one from the preparser file does not (also no hint that the rest of `preparser.ts` gets executed).

**Environment**

- Slidev version: 52.10.1
- Browser: Chrome
- OS: Mac

kaigritun:
Fixes #2402

## Problem

The `setup/preparser.ts` file in themes was being ignored while other setup files like `shiki.ts` were correctly loaded.

## Root Cause

In `packages/slidev/node/setups/preparser.ts`, the roots array only included:
- Addon roots
- User root

But NOT theme roots. Compare with how `setupShiki()` receives `resolved.roots` which includes themes.

## Solution

Resolve the theme root from headmatter (using the same logic as `options.ts`) and include it in the roots array passed to `loadSetups()`.

## Testing

- All 101 existing tests pass
- ESLint passes

## Changes

```diff
import { resolveAddons } from '../integrations/addons'
+import { resolveTheme } from '../integrations/themes'
import { getRoots } from '../resolver'
import { loadSetups } from './load'

export default function setupPreparser() {
  injectPreparserExtensionLoader(async (headmatter, filepath, mode) => {
    const addons = Array.isArray(headmatter?.addons) ? headmatter.addons : []

+   // Resolve theme from headmatter (matching logic from options.ts)
+   let themeRaw = headmatter?.theme as string | null | undefined
+   themeRaw = themeRaw === null ? 'none' : (themeRaw || 'default')

    const { userRoot } = await getRoots()
+   const [, themeRoot] = await resolveTheme(themeRaw, filepath)
+   const themeRoots = themeRoot ? [themeRoot] : []

    const roots = uniq([
+     ...themeRoots,
      ...await resolveAddons(addons),
      userRoot,
    ])
```

### DIFF
--- a/packages/slidev/node/setups/preparser.ts
+++ b/packages/slidev/node/setups/preparser.ts
@@ -2,6 +2,7 @@ import type { PreparserSetup } from '@slidev/types'
 import { uniq } from '@antfu/utils'
 import { injectPreparserExtensionLoader } from '@slidev/parser/fs'
 import { resolveAddons } from '../integrations/addons'
+import { resolveTheme } from '../integrations/themes'
 import { getRoots } from '../resolver'
 import { loadSetups } from './load'
 
@@ -10,8 +11,16 @@ export default function setupPreparser() {
     // Ensure addons is an array or an empty array if undefined
     const addons = Array.isArray(headmatter?.addons) ? headmatter.addons as string[] : []
 
+    // Resolve theme from headmatter (matching logic from options.ts)
+    let themeRaw = headmatter?.theme as string | null | undefined
+    themeRaw = themeRaw === null ? 'none' : (themeRaw || 'default')
+
     const { userRoot } = await getRoots()
+    const [, themeRoot] = await resolveTheme(themeRaw, filepath)
+    const themeRoots = themeRoot ? [themeRoot] : []
+
     const roots = uniq([
+      ...themeRoots,
       ...await resolveAddons(addons),
       userRoot,
     ])


antfu:
Why do you want to use a theme to change the preparser but not an addon? I believe this was on purpose to differentiate two integrations with different scope

kaigritun:
Good question! I based this on consistency with existing behavior - themes already support `setup/shiki.ts` and other setup files via `resolved.roots` (which includes themes), but preparser was the exception.

The original issue (#2402) reports that users expect parity: if `setup/shiki.ts` works in themes, `setup/preparser.ts` should too.

That said, if the intent is for preparser customization to be addon-only, I can close this PR - just wanted to match the pattern established by other setup files. Happy to follow whichever direction you prefer!

antfu:
Shiki is related to the theming, which is why it's included. We could update the docs to clear the behaviour instead.

jgosmann:
To me it was totally unexpected that `preparser.ts` is not loaded from the theme. Also I want to use the preparser in a theme for currently two things:

* Show a warning for deprecated keys in the theme config from the headmeatter.
* Extract and process slide headings as they require some special processing in the theme.

Splitting this into an add-on feels wrong because

* it is tightly coupled with the theme,
* requires to always configure the addon in addition to the theme.

kermanx:
> To me it was totally unexpected that `preparser.ts` is not loaded from the theme. Also I want to use the preparser in a theme for currently two things:
> 
> * Show a warning for deprecated keys in the theme config from the headmeatter.
> * Extract and process slide headings as they require some special processing in the theme.
> 
> Splitting this into an add-on feels wrong because
> 
> * it is tightly coupled with the theme,
> * requires to always configure the addon in addition to the theme.

I agree with you. It would be out of expectation if the preparser is not loaded from the theme. We may open a new PR to address this. However, `resolveTheme` and `resolveAddons` are not cheap operations, and `injectPreparserExtensionLoader` is not a good way to set preparsers. A refactor may be required to fix this.

jgosmann:
Is there anything I could do to help with this? I'm not very familiar with the code base and not sure what sort of refactoring you have in mind.

jgosmann:
It would be great if I could get some feedback on my PR; at least whether this is the right direction and there is a chance that this will merged. Otherwise, I'll have to split the preparser from my theme into a separate add-on.

PR #2458, opened 2026-02-10
chore: ignore ai config in gitignore

iiio2:
Suppose we're using skills. If we use skills it generally creates two folders with many files. I think we can ignore it in gitignore.


iiio2:
Thanks @antfu . :blush:

PR #2465, opened 2026-02-25
feat: add Mermaid renderer plugin

ysknsid25:
Slidev already can change Mermaid Diagram design using `theme` or `themeVariables`. It's great feature.

But if users try to 3rd Mermaid renderer libraries like [Beautiful Mermaid](https://github.com/lukilabs/beautiful-mermaid), they can't use it.

So I think I'd like to use 3rd Mermaid renderer libraries. It's easy.

# Slidev

- I added `mermaid-renderer` to `setupModules`
- Thereby, if there is a Custom Mermaid before an existing Mermaid, resolve it.

# User

- The user installs the Mermaid library they want to use. 
  - e.g.) `npm install beautiful-mermaid`
- The user creates `setup/mermaid-renderer.ts` in their directory.

```typescript
// setup/mermaid-renderer.ts
import { defineMermaidRendererSetup } from '@slidev/types'
import { renderMermaid } from 'beautiful-mermaid'

export default defineMermaidRendererSetup(() => {
  return (code, _options) => renderMermaid(code)
})
```

That's all.

# Check

Use existing Mermaid.

image

Use Beautiful Mermaid.

image

### DIFF
--- a/docs/.vitepress/customizations.ts
+++ b/docs/.vitepress/customizations.ts
@@ -44,6 +44,10 @@ export default [
     link: '/custom/config-mermaid',
   },
   {
+    text: 'Configure Mermaid Renderer',
+    link: '/custom/config-mermaid-renderer',
+  },
+  {
     text: 'Configure Routes',
     link: '/custom/config-routes',
   },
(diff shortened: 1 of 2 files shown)


antfu:
This is great! Would you also help to update docs about it?

ysknsid25:
Hi @antfu , long time no see :) 

> This is great! Would you also help to update docs about it?

Sure, sorry. I didn't notice. I'll add docs 🙏

ysknsid25:
@antfu 
I've added docs 🫡

antfu (approved):

PR #2467, opened 2026-02-27
fix(cli): avoid circular chunk warning between monaco types and shiki

mjvo:
This removes hard manual chunk assignment for Monaco types and Shiki in
`packages/slidev/node/vite/extendConfig.ts`.

Why:
- current manual assignments can produce:
  `Circular chunk: monaco/bundled-types -> modules/shiki -> monaco/bundled-types`

What changed:
- for Monaco types and Shiki ids, `manualChunks` now returns `undefined`
  (let Rollup decide chunk placement)
- existing icon and async module chunk rules remain unchanged

Validation:
- reproduced warning before change
- warning no longer appears after change
- local lint passes

before: circular chunk warning appears
after: warning no longer appears


antfu:
I think the reason we use manual chunking is that we want to pack the related code into a single chunk for better output structure and client load efficiency. In terms of the circular references, I think we should figure out why and where the shared chunk is to perform better manual chunking.

Removing it basically loses the original point of doing manual chunking, as both Monaco and Shiki generate too many chunks, which is suboptimal.

Other than improving the manual chunking, I think we may also add an option to disable manual chunking logic entirely if people prefer automatic chunking.

PR #2500, opened 2026-03-25
feat: Virtual Background for Web Camera

SentimentalK:
# Virtual Background for Web Camera

Add real-time virtual background support to Slidev's webcam overlay using **MediaPipe Selfie Segmentation**. 

### Key Features
- **4 Modes**: None (Raw), Blur Background (adjustable), Solid Color, and Custom Image.
- **Local File Support**: Upload local images (converted via `FileReader` to Data URLs) alongside remote URLs.
- **Persistence**: Remembers users' selection across reloads via local storage sync.

---

### Technical Implementation & Optimizations

We built a high-performance rendering pipeline on `<canvas>` ensuring smooth 30+ FPS even with browser-based AI inference.

#### 1. Lazy loading & Eager pre-loading
- The `@mediapipe/tasks-vision` WASM and structure initialize **only when the camera streams**.
- Deduplicated init promise handles multiple downstream triggers safely without double downloads.

#### 2. GPU Compositing (Zero `getImageData` on hot-path)
To maximize Frame Rate (FPS) and eliminate CPU overload:
- **Offscreen Canvas setup**: `bgCanvas` holds the background layer, `maskCanvas` holds the alpha transparency.
- **Alpha Masking Strategy**: We apply the segmentation mask onto the source video via `<canvas>` `destination-in` directly. 
- All graphics blending arithmetic is offloaded to the **GPU Canvas compositing** layer instead of synchronizing heavy `getImageData` back into the V8/JS loop per-frame.

#### 3. Motion-Adaptive Temporal Smoothing (EMA Buffer)
To reconcile the balance between **Suppressing Edge Flicker** and **Suppressing Edge Ghosting trailing effect**:
- **Pixel-level Diff delta tracking**: Calculates weights (`alpha`) per coordinate differential.
- Stable/Static silhouettes use `alpha=0.15` (heavy-damping smoothing filters out noise/ambient air flicker).
- Motion/Dynamic silhouette edges use `alpha=0.80` (fast-tracking tracks fast frame-swaps like head turns/hand gestures cleanly to eliminate residual light ghosts).

#### 4. Post-processing Alpha Feathering
Applied an anti-aliasing boundary smoothing:
- Before injecting output back to the DOM, the person silhouette is drawn with a subtle `ctx.filter = 'blur(1px)'` to blend naturally into higher contrast backgrounds without hard jagged edge artifacts.

---

### File Structure
- `logic/virtualBackground.ts` — Compositing engine + GPU masking logic.
- `internals/VirtualBackgroundSettings.vue` — UI Selectors & Inputs.
- Added `@mediapipe/tasks-vision` standard lookup to dependencies.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -90,6 +90,9 @@ catalogs:
     '@antfu/utils':
       specifier: ^9.3.0
       version: 9.3.0
+    '@mediapipe/tasks-vision':
+      specifier: ^0.10.18
+      version: 0.10.34
     '@shikijs/engine-javascript':
       specifier: ^4.0.1
       version: 4.0.1
@@ -736,7 +739,7 @@ importers:
         specifier: catalog:icons
         version: 1.2.4
       '@mediapipe/tasks-vision':
-        specifier: ^0.10.18
+        specifier: catalog:frontend
         version: 0.10.34
       '@shikijs/engine-javascript':
         specifier: catalog:frontend


SentimentalK:
If Slidev supports webcam features, I believe virtual background is an important capability rather than an optional add-on.

In terms of size, the overall node_modules is already around ~480MB, so an additional ~35MB dependency does not significantly change the overall footprint.

Also, `@mediapipe/tasks-vision` is not strictly required, but it provides much better performance and quality for background segmentation. This is similar to how Google Meet implements virtual background.

That said, I'm open to making this optional (e.g. lazy-loaded or behind a feature flag) if you think that would be a better trade-off. Without it, the initial load would be significantly slower, and users may experience noticeable lag during background processing.

kermanx:
Thanks for the info! In my view, Slidev is already quite bloated at ~480MB, so adding another 35MB is something we should avoid if possible. How about we dynamically import `@mediapipe/tasks-vision` (without the need for a flag)? We could fall back to a lower-performance mode and issue a warning if it's not installed—similar to how we handle `playwright-chromium`.

SentimentalK:
To be honest, I don’t think the lower-performance mode is useful here. It introduces noticeable lag, and I’d prefer not to include it in this case.

I also want to correct my earlier point about size — I don’t think 35MB is something we should be concerned about relative to 1.2GB. The 35MB is only relevant at install time and does not impact the production bundle. The WASM runtime (~33MB, which makes up most of the package) is loaded from `cdn.jsdelivr.net` at runtime (see [virtualBackground.ts #line 24](https://github.com/SentimentalK/slidev/blob/fc53d1b616e1d94779f4b91c55e05b2432079c42/packages/client/logic/virtualBackground.ts#L24)), rather than being bundled by Vite.
```
sentimentalk@SentimentalKs-MacBook-Pro slidev % du -sh node_modules/.pnpm
1.2G	node_modules/.pnpm
```
If size is still a concern, I think the next step would be to focus on cleanup and optimization instead.

antfu:
If you're using mac, it already has a system-wide built-in virtual background, so we don't need to handle anything there. As for other os, I'm not entirely sure, but if we want to support something like that, I think it's better to provide it as a hook. This would allow the community to build plugins for additional effects beyond just virtual backgrounds.

In my opinion, it's better to keep this separate. I'm open to seeing proposals for implementing those hooks, but regarding this specific PR:

1. We shouldn't have it built-in.
2. Keeping it separate is better for bundle size.
3. It offers more flexibility for the community to implement and innovate.

Thanks a lot for the PR. I think this is definitely a very interesting aspect for Slidev to improve. I'm looking forward to seeing your PR if you want to look into how we can design an API for add-ons to hook into the streaming process.

PR #2510, opened 2026-03-27
feat: add laser pointer feature

### LINKED ISSUE #2509: The laser pointer function (like the ms PPT)
Now we can draw and mark to indicate to the reader, but the laser pointer function of PPT is more convenient and eye-catching. It would be even better if this function could be added in slidev.

hzb666:
resolves #2509

  ## Summary

  Add a laser pointer mode to Slidev presentation controls.

  ## What changed

  - add a dedicated laser pointer toggle to navigation controls
  - add a matching context menu action for enabling and disabling laser pointer mode
  - introduce a reusable laser pointer icon component
  - extend shared cursor state to support both `cursor` and `laser` styles
  - render laser pointer feedback in both presenter and play views
  - hide the native cursor over slide content while laser pointer mode is active
  - make laser pointer mode and presenter cursor mode mutually exclusive
  - persist laser pointer preference in local storage

  ## Implementation notes

  - `showLaserPointer` is stored alongside presenter cursor preferences
  - presenter view syncs cursor position through shared state and switches style based on the active mode
  - play view can render the local laser pointer and also falls back to shared cursor state when needed

image
image
image
image


kermanx (approved):
Thank you for the PR~

I've made some updates to it.

antfu:
Thanks for the feature - one thing I'm a little bit concerned about is that the icon is a little bit confusing with the recording button. Do you think we can have a better alternative for showing that this is actually a laser pointer?

Maybe `i-carbon-magic-wand-filled`

CleanShot 2026-04-08 at 10 59 12@2x

hzb666:
> Thanks for the feature - one thing I'm a little bit concerned about is that the icon is a little bit confusing with the recording button. Do you think we can have a better alternative for showing that this is actually a laser pointer?
> 
> Maybe `i-carbon-magic-wand-filled`
> 
> CleanShot 2026-04-08 at 10 59 12@2x

An icon in www.iconfont.cn is quite suitable and i created a similar one and did some editings on my own to avoid any licence conflicts. auth：[524yangyang](https://www.iconfont.cn/user/detail?uid=847075&nid=GbkEECDKYLCl)

</img>

kermanx (approved):

antfu:
@copilot resolve the merge conflicts in this pull request

antfu:
CleanShot 2026-04-28 at 11 43 25@2x
I have moved the entry to the settings

Issue #2439, opened 2026-01-20
Custom style.css must use !important to overwrite default style


game1024:
Slidev  v52.11.3

**Describe the bug**

I want to set custom style for vclick target in style.css, but it not work, I must use !important to overwrite default style
my slides.md
```
---
mdc: true
---
# title1

<v-clicks every="2">

- Item 1.1
- Item 1.2
- Item 2.1
- Item 2.2

</v-clicks>
```

my style.css
```css
.slidev-vclick-target {
  transition: all 500ms ease;
}

.slidev-vclick-hidden {
  opacity: 1;
  transform: scale(0,0)
}
```

<img width="2818" height="1624" alt="Image" src="https://github.com/user-attachments/assets/e6e7f56f-39b7-48a7-94d3-d217ef1ca87f" />

**Minimal reproduction**

1. set style.css
2. run `bun run dev`
3. Click on '....'
**Environment**

- Slidev version: v52.11.3
- Browser: Microsoft Edge 
- OS: macos

If you are using Slidev globally (i.e. `npx slidev` or `npm i -g slidev`), please try to reproduce the issue in a local project (i.e. `npm create slidev@latest`).

harrybin:
we have the same poblem with a custom theme. Since one of the last versions also the css vars, especially the `--slidev-theme-primary` is no longer used without `!important`.
It looks like the slidev internal `vars.css` is loaded after the one in the theme and project style.ss.

Issue #2412, opened 2025-12-20
Has the priority order of theme styles changed in version 52.10.0?


ras0q:
**Describe the bug**

Previously, setting `--slidev-theme-primary` in the theme or custom `styles.css` file would override the default.
However, starting with version 52.11.1, the default setting now takes precedence.

```css:styles.css
:root {
  --slidev-theme-primary: #ff0000;
}
```

|>=52.9.1|<=52.10.0|
|-|-|
|<img width="440" height="302" alt="Image" src="https://github.com/user-attachments/assets/a5e86b8c-9843-4dd3-874b-3b1decb74580" />|<img width="515" height="319" alt="Image" src="https://github.com/user-attachments/assets/b62f8594-1c6d-4bbc-822a-416488ba8e18" />|

**Minimal reproduction**

1. Open the following Stackblitz projects.
2. Run `pnpm install && npm run dev` (automatically).
3. See the page 2.
4. In previous versions, the custom setting #ff0000 was set as the color, while in the new version, the default #3ab9d5 is set.

Stackblitz reprosductions:

- `"@slidev/cli": "52.9.1"`: https://stackblitz.com/edit/github-3etyxpup?file=package.json
- `"@slidev/cli": "52.10.0"`: https://stackblitz.com/edit/github-3etyxpup-xjzwjtt3?file=package.json

**Environment**

- Slidev version: 52.10.0
- Browser: Microsoft Edge
- OS: Windows 11

31ch:
I had a quick look at the problem. The root cause seems to be commit 593b9f8ebec6e196b2081ee614bbeee6859a0505, probably the changes within `packages/slidev/node/virtual/styles.ts`. The changed import handling somehow seems to break the CSS cascade.

---

🚀 **Workaround for custom themes:**

If we change our `styles/layout.css` color definitions from `:root` to `:root:root` custom colors are working again:

```css
:root {
  --slidev-theme-primary: #fa9400;
  --slidev-theme-secondary: #424242;
  --slidev-theme-tertiary: #ffffff;
}
```

```css
:root:root {
  --slidev-theme-primary: #fa9400;
  --slidev-theme-secondary: #424242;
  --slidev-theme-tertiary: #ffffff;
}
```

---

@antfu  would be great if you could have a look at this issue as its literally breaking all custom themes using the styles from `@slidev/client/styles/layouts-base.css` with custom colors as specified above!

Btw: this issue is most possibly heavily related to (or duplicate of):
[...]

PR #2511, opened 2026-03-29
fix(styles): use direct imports to ensure correct CSS order

### LINKED ISSUE #2495: Styles are appllied in incorrect order
**Describe the bug**

I have a `styles.css` with the following content:

```
h2 {
    font-size: 2rem;
}
```

Up to and including Slidev 52.9.1, this took effect as expected on slides with `layout: none`. (Other layouts, at least with the default theme, have a `.slidev-layout h2` rule with higher specificity.) But since version 52.10.0 (and still with version 52.14.1), another rule with `font-size: inherit` takes precedence.

Looking at the computed CSS in the dev tools in version 52.9.1 shows this:

Image

With version 52.10.0 this changed to:

Image

Note that the order of `2rem` and `inherit` is swapped.

I suspect that his is related to:

#2380, #2412, #2439, #2457

**Minimal reproduction**

[I created a minimal reproduction in jgosmann/slidev-styles-bug.](https://github.com/jgosmann/slidev-styles-bug). It is a freshly created Slidev presentation. The second slide shows the problem. It has an h2-heading that should have the `font-size` styled with `styles.css`, but the style is overridden with a default.

**Environment**

- Slidev version: 52.10.0 and 52.14.1
- Browser: Firefox 148.0.2
- OS: macOS Tahoe 26.3.1 (a)

cyphercodes:
## Summary

Replace `import.meta.glob` with direct imports for CSS files to ensure styles are loaded in the correct order.

## Problem

Since v52.10.0, user styles in `styles.css` are being overridden by default styles because `import.meta.glob` does not guarantee CSS import order (fixes #2495).

## Solution

Restore the behavior from before commit 593b9f8e where direct imports were used. This ensures:
1. Reset and UnoCSS preflights load first
2. Default client styles load next  
3. User styles load after defaults (allowing overrides)
4. Feature-specific styles (KaTeX, Shiki) load after user styles
5. UnoCSS utilities load last

## Changes

- Replaced `makeAbsoluteImportGlob` with direct `import` statements for client and user styles
- Added back `existsSync` check for user styles to maintain the same file discovery logic
- Preserved the order: reset → defaults → user → features → uno.css

## Related Issues

- Fixes #2495
- Related to #2380, #2412, #2439, #2457

### DIFF
--- a/packages/slidev/node/virtual/styles.ts
+++ b/packages/slidev/node/virtual/styles.ts
@@ -1,4 +1,5 @@
 import type { VirtualModuleTemplate } from './types'
+import { existsSync } from 'node:fs'
 import { join } from 'node:path'
 import { resolveImportUrl, toAtFS } from '../resolver'
 import { makeAbsoluteImportGlob } from '../utils'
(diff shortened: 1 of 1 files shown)


kermanx:
Thank you for the PR~

However, there is a reason to use `makeAbsoluteImportGlob` over statically checking the existence of the files: `import.meta.glob` handles HMR automatically (#2380). However, `import.meta.glob` imports are always moved by Vite to the top of the file; that's why #2457 was proposed.

I've proposed #2512 to fix this issue while preserving `makeAbsoluteImportGlob`, hopefully it works.

PR #2278, opened 2025-09-16
feat: handout export (slides + notes + cover/footer/ending)

oripka:
Introduces a **dedicated handout export** that renders one **A4 page per slide** with:

* a scaled slide preview,
* speaker notes,
* and a customizable footer.

Optionally, a **multi-page cover** can be prepended.
Export is supported both in the **CLI** and the **browser exporter**, designed for reliable print output.

### Handout rendering and routes

* New print routes for **handout** and **cover**, enabled whenever print or browser exporter features are active.

  * `packages/client/setup/routes.ts`
  * `packages/client/pages/handout/print.vue`
  * `packages/client/pages/cover/print.vue`

### Page layout

* Each page: **slide (top)**, **notes (middle)**, **footer (bottom)**.
* Uses fixed A4 dimensions in CSS pixels (96 DPI).
* Enforces print media styles for consistency.

  * `packages/client/internals/PrintHandout.vue`
  * `packages/client/internals/PrintContainerHandout.vue`

### Virtual global components (user-overridable)

* Auto-discovers `handout-cover.vue` / `HandoutCover.vue` and `handout-bottom.vue` / `HandoutBottom.vue`.
* Falls back to no-op components if not present.
* Props like `pageNumber` are forwarded (supports camel/kebab case).

  * `packages/slidev/node/virtual/handout-components.ts`
  * `packages/slidev/node/virtual/index.ts`

### CLI integration

* `slidev export --handout` → `<output>-handout.pdf`.
* `--cover` prepends cover pages if defined.
* `--range` limits exported slides.
* Uses **fixed A4 viewport** and **print media** for stability.

  * `packages/slidev/node/commands/export.ts`
  * `packages/slidev/node/cli.ts`
  * `packages/types/src/cli.ts`

### Print safety

* Disables slide-sized `@page` rules for handout/cover routes so A4 applies cleanly.

  * `packages/client/composables/usePrintStyles.ts`

### Navigation updates

* `isPrintMode` now includes `handout` and `cover`.

  * `packages/client/composables/useNav.ts`

### Starter examples

* Demo components provided:

  * `demo/starter/handout-cover.vue`
  * `demo/starter/handout-bottom.vue`

### CLI

```bash
# Regular slides export (unchanged)
slidev export --output slides.pdf

# Handout export
slidev export --handout --output slides.pdf

# Handout with cover
slidev export --handout --cover --output slides.pdf

# With range
slidev export --handout --range 1,3-5 --output slides.pdf
```

[...]

PR #2515, opened 2026-04-02
feat: Add ability to print the overview

jzarzeckis:
I have slides with very extensive extra reading material in the presenter notes - and it's such a pity that I can't share the overview with my colleagues as a learning material. This adds ability to export `/overview`

Here's what the export PDF looks like:
[slides-export.pdf](https://github.com/user-attachments/files/26435273/slides-export.pdf)

### DIFF
--- a/docs/guide/exporting.md
+++ b/docs/guide/exporting.md
@@ -199,6 +199,16 @@ You can generate the PDF outline by passing the `--with-toc` option:
 $ slidev export --with-toc
 ```
 
+### Overview
+
+You can export the overview page (slides with presenter notes side by side) as an A4 portrait PDF by passing the `--overview` option:
+
+```bash
+$ slidev export --overview
+```
+
+This is useful when you want a printable summary of your entire presentation including all speaker notes. Each slide is rendered alongside its notes, and slides are never split across page boundaries.
+
 ### Omit Background
 
 When exporting to PNGs, you can remove the default browser background by passing `--omit-background`:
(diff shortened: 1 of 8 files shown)


kermanx:
> https://github.com/slidevjs/slidev/pull/2278#issuecomment-4088260219:
>
> I personally think it could be better if we could support any custom print template, just like https://github.com/slidevjs/slidev/pull/1513. Then this export layout can be implemented as a built-in print template, alongside the default one.

For now, we can use packages generated by pkg.pr.new for this PR: https://github.com/slidevjs/slidev/pull/2515#issuecomment-4182207486

jzarzeckis:
> > [#2278 (comment)](https://github.com/slidevjs/slidev/pull/2278#issuecomment-4088260219):
> > I personally think it could be better if we could support any custom print template, just like #1513. Then this export layout can be implemented as a built-in print template, alongside the default one.
> 
> For now, we can use packages generated by pkg.pr.new for this PR: [#2515 (comment)](https://github.com/slidevjs/slidev/pull/2515#issuecomment-4182207486)

@kermanx Thanks for the tip!

Fair point about different "Printout Templates". Then again - honestly I had missed #2278 exists - and it already covers the problem I was hoping to solve (I believe we can then close this PR as a duplicate of #2278). I don't think realistically there would be more PDF template types beyond the 2 (clean slides; handout) options - it is the basic functionality students expect from powerpoint and similar.

PR #2521, opened 2026-04-06
docs: add slidev-agent addon information

christian-bromann:
Added slidev-agent addon with details for integration.


antfu:
Looks awesome! Thank you for building it!

PR #2540, opened 2026-04-14
feat: onSlideEnter/Leave after mounted + pass idx

maxkurze1:
This PR refactors the `onSlideEnter` and `onSlideLeave` functions providing 3 contributions

### 1. replace `watchEffect` with `watch` 

Using `watchEffect` for the definition of `onSlideEnter` would cause unintended invocations triggered by arbitrary references used inside the callback. I would consider this a bug in the previous implementation, which was fixed by replacing it with a plain `watch` which only triggers based on the explicitly provided dependencies.

To retain the initial invocation on slide load I configured `{ immediate: true }`, though I think it is not necessary anymore since the hook should always be set up before mounting and thus will be triggered by `ismouted.value = true` anyhow.

### 2. make sure `onSlideEnter` is (always) called after `onMounted`

AFAICT this guarantee only provides benefits. I would argue that most of the use cases of `onSlideEnter`/`onSlideLeave` want to do some setup/DOM configurations and are thus expecting this ordering already, even though it is currently not guaranteed.

The new version of `onSlideEnter` uses a dependency on `ismounted` to get as close as possible to the desired ordering. However, I am still not entirely sure whether the ordering is actually guaranteed now, as the Vue documentation is not as explicit as I would like it to be.

Maybe one should consider adding `nextTick`? But I'd argue that it is already good enough.

### 3. passing the current / previous slide number to `onSlideEnter`/`onSlideLeave`

In my personal developments I found these arguments to be quite handy, so I decided to include them here as well. Sure, one could argue that one of the parameters is useless as it just mirrors `$page.value`, but that didn't really matter to me.

### DIFF
--- a/packages/client/logic/slides.ts
+++ b/packages/client/logic/slides.ts
@@ -47,6 +47,6 @@ export function onSlideLeave(cb: (to: number, from: number | undefined) => any) 
     watch(() => $nav.value.currentSlideNo, (to, from) => {
       if ($page.value === from)
         cb(to, from)
-    }, { immediate: true })
+    })
   })
 }
--- a/packages/client/logic/slides.ts
+++ b/packages/client/logic/slides.ts
@@ -1,6 +1,6 @@
 import type { SlideRoute } from '@slidev/types'
 import { slides } from '#slidev/slides'
-import { computed, watch, watchEffect } from 'vue'
+import { computed, onMounted, onUnmounted, ref, watch } from 'vue'
 import { useSlideContext } from '../context'
 
 export { slides }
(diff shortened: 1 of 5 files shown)


kermanx (approved):

packages/client/logic/slides.ts
```diff
@@ -28,30 +29,24 @@ export function useIsSlideActive() {
...
 export function onSlideEnter(cb: (to: number, from: number | undefined) => any) {
-  const ismounted = useIsMounted()
   const { $page, $nav } = useSlideContext()
 
-  // Note: using watch + immediate rather than watchEffect since the latter could have
-  //   unexpected reloads triggered by references used in `cb`
-  // Note: using `ismounted` to make sure `onSliderEnter` is only called after `onMounted`
-  watch(() => [$nav.value.currentSlideNo, ismounted.value], ([to, mounted], from) =>
-    $page.value === to && mounted && cb(to, from ? from[0] as number : undefined), { immediate: true, flush: 'post' })
+  tryOnMounted(() => {
```
maxkurze1, line 35: Using on `tryOnMounted` or even the normal `onMounted` seems more clean to me, I also considered this. The problem being that if you have a slide with both `onMounted` and `onSlideEnter` on it, their order isn't guaranteed. In this case, as `onSlideEnter` basically just desugars to `onMounted`, their ordering in the file determines their invocation order.

E.g. the following code invokes `onSlideEnter` first and only after it `onMounted`:

```vue
<script setup lang="ts">
import { onMounted } from "vue"
import { onSlideEnter, onSlideLeave } from "@slidev/client";

onSlideEnter(() => {
  console.log("enter")
})
onMounted(() => {
  console.log("mount")
})
</script>
```

I've instead tried, the best I could, to make sure that `onSliderEnter` is always called **after** `onMounted`. In my implementation that was caused by the `ismounted` value which was updated during the mount but for some Vue internal reason it seemed to trigger the watch hook only after all mounts executed.

Unfortunately, I couldn't find a cleaner / more reliable way to force this ordering. I also didn't have the time yet to look into Vue and verify if this approach actually works out all the time. My intuition was that it seemed reasonable that the watch hook is only checked/triggered after another "internal tick" or something like this and during testing I never observed a different behavior.

PR #2541, opened 2026-04-14
Fix scaling in export view

maxkurze1:
This fixes a visual bug in the export view. As the export view does not use the same `SlideWrapper` component the scaling is not provided correctly. This can be notices at the rough notations for instance:

---

Screenshot from 2026-04-14 20-19-40
Screenshot from 2026-04-14 20-20-40

---

By adding the container `--slidev-slide-container-scale` the scale is provided during export render fixing the visuals:

---

Screenshot from 2026-04-14 20-20-17
Screenshot from 2026-04-14 20-20-27

---

### DIFF
--- a/packages/client/pages/export.vue
+++ b/packages/client/pages/export.vue
@@ -351,6 +351,10 @@ h2 {
 #export-content {
   --uno: pointer-events-none;
 }
+
+#export-container {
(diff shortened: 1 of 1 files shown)


kermanx (approved):
Thanks~

PR #2543, opened 2026-04-15
Replace pnpm with npm in README.md

### LINKED ISSUE #2516: README in scaffolded project always references pnpm regardless of chosen package manager
When running **bun create slidev** (or npm create slidev, etc.) and choosing a package manager other than pnpm during the "Install and start it now?" prompt, the **README.md** written into the new project still contains hardcoded **pnpm** commands.

**Steps to reproduce**
1. Run bun create slidev
2. Enter a project name
3. When asked "Install and start it now?", answer Yes
4. When asked "Choose the package manager", select bun (or any non-pnpm option)
5. Open the README.md in the newly created project

**Expected behavior**
The README.md should reflect the selected package manager. For example, when bun is chosen:
- `bun install`
- `bun dev`

**Actual behavior**
The README.md always contains:
- `pnpm install`
- `pnpm dev`

**Root cause**
packages/create-app/template/README.md is a static file that hardcodes pnpm. In packages/create-app/index.mjs, the file is copied verbatim at lines 87–89 before the package manager is selected at line 116–121. The chosen agent is only used to run the install/dev commands, never to update the README.

**Suggested fix**
Defer writing the README until after the package manager is known, and substitute the agent value into the README content dynamically — similar to how package.json receives the project name via write('package.json', JSON.stringify(pkg, null, 2)).

**A secondary issue:**
When the user declines to install immediately, the auto-detection on lines 97–99 (pkgManager) only handles pnpm, yarn, and npm — bun and deno are not detected. This affects the console hints printed in the else branch (lines 134–135), but a decision should be made about what to use in the README in this case (e.g., keep pnpm as default, or prompt for the package manager regardless).

**Environment**
- Slidev version: 52.14.1 (create-app)
- Browser: N/A (issue is in the CLI scaffolding, not the browser)
- OS: macOS 26.4

hrithik18k:
closes #2516

### DIFF
--- a/packages/create-app/template/README.md
+++ b/packages/create-app/template/README.md
@@ -2,8 +2,8 @@
 
 To start the slide show:
 
-- `pnpm install`
-- `pnpm dev`
+- `npm install`
+- `npm run dev`
 - visit <http://localhost:3030>
 
 Edit the [slides.md](./slides.md) to see the changes.


hrithik18k:
@antfu review please .

maxkurze1:
The whole point of the issue was that the package manager **should not** be hard-coded in the `README.md`, just changing the hard-coded name from `pnpm` to `npm` does not solve the issue!

antfu:
At this point in time, I see no point in using npm over pnpm. We are recommending pnpm to be the package manager for users; for those who don't want to use it, they can always change the command on their own.

PR #2545, opened 2026-04-15
feat(create-app): automatically detect and substitute package manager

### LINKED ISSUE #2516: README in scaffolded project always references pnpm regardless of chosen package manager

hrithik18k:
closes #2516

### DIFF
--- a/packages/create-app/index.mjs
+++ b/packages/create-app/index.mjs
@@ -79,12 +79,9 @@ async function init() {
   const templateDir = path.join(__dirname, 'template')
 
   const write = (file, content) => {
-    const targetPath = renameFiles[file]
-      ? path.join(root, renameFiles[file])
-      : path.join(root, file)
+    const targetPath = path.join(root, renameFiles[file] ?? file)
     if (content)
       fs.writeFileSync(targetPath, content)
-
     else
       copy(path.join(templateDir, file), targetPath)
   }
@@ -99,24 +96,23 @@ async function init() {
 
   write('package.json', JSON.stringify(pkg, null, 2))
 
+  console.log(green('  Done.\n'))
+
   function getPkgManager() {
+    const pm = []
     if (typeof Deno !== 'undefined')
-      return 'deno'
+      pm.push('deno')
     if (typeof Bun !== 'undefined')
-      return 'bun'
+      pm.push('bun')
     const userAgent = process.env.npm_config_user_agent || ''
     const execPath = process.env.npm_execpath || ''
     if (execPath.includes('pnpm') || userAgent.includes('pnpm'))
-      return 'pnpm'
+      pm.push('pnpm')
     if (execPath.includes('yarn') || userAgent.includes('yarn'))
-      return 'yarn'
-    return 'npm'
+      pm.push('yarn')
+    return pm.length === 1 ? pm[0] : null
   }
   const pkgManager = getPkgManager()
-
-  const related = path.relative(cwd, root)
-
-  console.log(green('  Done.\n'))
 
   /**
    * @type {{ yes: boolean }}
@@ -125,16 +121,16 @@ async function init() {
     type: 'confirm',
     name: 'yes',
     initial: 'Y',
-    message: 'Install and start it now?',
+    message: `Install and start it now${pkgManager ? ` using ${pkgManager}` : ''}?`,
   })
 
   if (yes) {
-    const { agent } = await prompts({
+    const agent = pkgManager || (await prompts({
       name: 'agent',
       type: 'select',
       message: 'Choose the package manager',
       choices: ['npm', 'yarn', 'pnpm', 'bun', 'deno'].map(i => ({ value: i, title: i })),
-    })
+    }).agent)
 
     if (!agent)
       return
@@ -147,7 +143,7 @@ async function init() {
     writeReadme(pkgManager)
     console.log(dim('\n  start it later by:\n'))
     if (root !== cwd)
-      console.log(blue(`  cd ${bold(related)}`))
+      console.log(blue(`  cd ${bold(path.relative(cwd, root))}`))
 
     console.log(blue(`  ${pkgManager} install`))
     console.log(blue(`  ${pkgManager} run dev`))
@@ -156,11 +152,11 @@ async function init() {
     console.log()
   }
 
-  function writeReadme(pm) {
+  function writeReadme(pm = 'npm') {
     const readmeTemplate = fs.readFileSync(path.join(templateDir, 'README.md'), 'utf-8')
     const readmeContent = readmeTemplate
-      .replace('{{PM_INSTALL}}', `${pm} install`)
-      .replace('{{PM_DEV}}', `${pm} run dev`)
+      .replace('npm install', `${pm} install`)
+      .replace('npm run dev', `${pm} run dev`)
     write('README.md', readmeContent)
   }
 }
@@ -169,7 +165,6 @@ function copy(src, dest) {
   const stat = fs.statSync(src)
   if (stat.isDirectory())
     copyDir(src, dest)
-
   else
     fs.copyFileSync(src, dest)
 }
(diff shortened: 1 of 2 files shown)


kermanx:
Hello! Thanks for your work on this. I noticed that another similar PR was opened for the same issue: #2544. You might want to check it out.

packages/create-app/index.mjs
```diff
@@ -101,9 +103,17 @@ async function init() {
...
+  const pkgManager = (RE_PNPM.test(execPath) || RE_PNPM.test(userAgent))
     ? 'pnpm'
-    : RE_YARN.test(process.env.npm_execpath || '') ? 'yarn' : 'npm'
+    : RE_YARN.test(execPath) || RE_YARN.test(userAgent)
+      ? 'yarn'
+      : RE_BUN.test(execPath) || RE_BUN.test(userAgent)
+        ? 'bun'
+        : RE_DENO.test(execPath) || RE_DENO.test(userAgent)
+          ? 'deno'
+          : 'npm'
```
maxkurze1: i think it makes sense to check

```
  if (typeof Deno !== 'undefined') return 'deno';
  if (typeof Bun !== 'undefined') return 'bun';
```

as well AFAICT deno does not set the useragent - btw. did you test with all the package managers?

packages/create-app/index.mjs
```diff
@@ -130,10 +140,12 @@ async function init() {
     if (!agent)
       return
 
+    writeReadme(agent)
```
maxkurze1, line 138: i don't see why you write the readme so late - i would rather move it just below writing all the other files line 98

packages/create-app/index.mjs
```diff
@@ -144,6 +156,16 @@ async function init() {
     console.log(`  ${cyan('●')} ${blue('■')} ${yellow('▲')}`)
     console.log()
   }
+
+  function writeReadme(pm) {
+    const readmeTemplate = fs.readFileSync(path.join(templateDir, 'README.md'), 'utf-8')
+    const installCmd = pm === 'yarn' ? 'yarn' : `${pm} install`
```
maxkurze1: doesn't yarn have an `yarn install` as well?

packages/create-app/index.mjs
```diff
@@ -144,6 +156,16 @@ async function init() {
...
+    const devCmd = pm === 'yarn' ? 'yarn dev' : `${pm} run dev`
```
maxkurze1: yarn should also support `yarn run dev`

packages/create-app/index.mjs
maxkurze1: same comment about yarn here as well

packages/create-app/index.mjs
hrithik18k: Done — added typeof Deno !== 'undefined' and typeof Bun !== 'undefined' checks, and removed the RE_BUN/RE_DENO regex since they're no longer needed

packages/create-app/index.mjs
hrithik18k: The README is written as early as possible — right after the package manager is determined. In the "yes" path it must wait for user selection; in the "no" path it's written immediately

packages/create-app/index.mjs
hrithik18k: Fixed — removed yarn special-casing. Now all PMs use ${pm} install uniformly

packages/create-app/index.mjs
hrithik18k: Fixed — all PMs now use ${pm} run dev uniformly, including the console hints

packages/create-app/index.mjs
hrithik18k: Fixed! Removed yarn special-casing from the console hints as well. All PMs now use ${pm} install / ${pm} run dev uniformly.

hrithik18k:
Thanks for the review and the heads up @kermanx and @maxkurze1!

I've pushed an update addressing all the review feedback:

Added typeof Deno !== 'undefined' and typeof Bun !== 'undefined' checks for runtime detection, and removed the unused RE_BUN/RE_DENO regex fallback.
Removed yarn special-casing since it fully supports standard yarn install and yarn run dev commands.
Verified that writeReadme() runs as early as possible (immediately after the package manager is determined in both the automatic and manual selection paths).
Testing: I have manually tested the generator script locally on my end. The dynamic scaffolding works perfectly. For example, selecting bun during scaffolding successfully generates a README.md with - bun install and - bun run dev, and the console hints also map correctly.

Let me know if there's anything else needed here!

hrithik18k:
Hey @kermanx — just a heads up that #2544 is still a draft with incomplete TODOs, while this PR fully addresses the issue including the README template substitution. Happy to get this reviewed whenever you have a moment!

packages/create-app/index.mjs
```diff
@@ -101,9 +101,17 @@ async function init() {
...
+  const execPath = process.env.npm_execpath || ''
+  const pkgManager = typeof Deno !== 'undefined'
+    ? 'deno'
+    : typeof Bun !== 'undefined'
+      ? 'bun'
+      : (RE_PNPM.test(execPath) || RE_PNPM.test(userAgent))
+          ? 'pnpm'
+          : RE_YARN.test(execPath) || RE_YARN.test(userAgent)
+            ? 'yarn'
+            : 'npm'
```
kermanx:
```suggestion
function getPkgManager() {
  if (typeof Deno !== 'undefined')
    return 'deno'
  if (typeof Bun !== 'undefined')
    return 'bun'
  const userAgent = process.env.npm_config_user_agent || '';
  const execPath = process.env.npm_execpath || '';
  if (execPath.includes('pnpm') || userAgent.includes('pnpm'))
    return 'pnpm'
  if (execPath.includes('yarn') || userAgent.includes('yarn'))
    return 'yarn'
  return 'npm'
}

const pkgManager = getPkgManager()
```

kermanx (changes requested):

packages/create-app/index.mjs
hrithik18k: Done! Addressed in the latest commit e48613f. Thanks for the suggestion!

hrithik18k:
Hey @kermanx — I've addressed your feedback in the latest commit (e48613f). Extracted getPkgManager() as a named function and simplified the detection logic using .includes() as suggested. Ready for re-review whenever you have a moment!

kermanx (approved):
Thanks to @hrithik18k and @maxkurze1 for implementing this feature!

PR #2551, opened 2026-04-18
fix(client): make drawing stroke-width control clickable

### LINKED ISSUE #2550: Clicking the drawing stroke-width control does nothing on desktop and touch devices
### Describe the bug

The drawing toolbar's "Adjust stroke width" control is not actually clickable/tappable.

On current `main`, that control is implemented with a floating-vue `Menu`, which is hover-oriented. In practice, clicking the button does nothing on desktop, and tapping it does nothing on touch devices.

### Reproduction

1. Open a Slidev deck in presenter mode or another view with the drawing toolbar available.
2. Enable drawing.
3. Click/tap the "Adjust stroke width" button.

### Actual behavior

- Desktop: clicking the button does nothing.
- Touch/mobile: tapping the button does nothing.
- On pointer devices, the control is effectively hover-based rather than click-based.

### Expected behavior

The stroke-width control should open as an interactive popover when clicked/tapped, and the slider should remain usable.

### Additional context

I have a one-file patch ready in `packages/client/internals/DrawingControls.vue` that replaces the current hover-oriented interaction with a click/touch popover.

enieuwy:
Closes #2550.

## Summary

This makes the drawing toolbar's stroke-width control open as an interactive popover when clicked/tapped.

On current `main`, the control is implemented with floating-vue `Menu`, which is hover-oriented. In practice, clicking the button does nothing on desktop, and tapping it does nothing on touch devices.

## What changed

- replace the current hover/menu-based stroke-width control with a small local popover
- open it on click/tap
- render the popover and backdrop via `Teleport` so they remain interactive above the toolbar
- close it when clicking the backdrop or after changing the slider

## Scope

This PR intentionally stays to a single file:

- `packages/client/internals/DrawingControls.vue`

It only fixes the stroke-width control interaction bug and does not change drawing behavior itself.

## Verification

- `pnpm -r --filter="./packages/**" --parallel run build`
- verified locally that the control:
  - opens on desktop click
  - opens on mobile/touch tap
  - allows interacting with the slider
  - closes via the backdrop

### DIFF
--- a/packages/client/internals/DrawingControls.vue
+++ b/packages/client/internals/DrawingControls.vue
@@ -1,5 +1,5 @@
 <script setup lang="ts">
-import { Menu } from 'floating-vue'
+import { computed, ref } from 'vue'
 import { useDrawings } from '../composables/useDrawings'
 import Draggable from './Draggable.vue'
 import IconButton from './IconButton.vue'
@@ -18,6 +18,15 @@ const {
   brushColors,
 } = useDrawings()
 
+const strokePopoverOpen = ref(false)
+const strokeBtnRef = ref<HTMLElement>()
+const strokePopoverStyle = computed(() => {
+  const el = strokeBtnRef.value
+  if (!el)
+    return {}
+  const rect = el.getBoundingClientRect()
+  return { top: `${rect.bottom + 4}px`, left: `${rect.left}px` }
+})
 function undo() {
   drauu.undo()
 }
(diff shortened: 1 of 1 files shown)


antfu:
I don't think we should introduce a custom way of doing floating panels other than floating-vue. floating-vue does provided click based trigger, if you think there is room to improve the UX, we should update the options of floating-vue instead of duplicating it with teleport (which would introduce other problems and would be hard to maintain). Feel free to open an new PR. Thanks.

PR #2553, opened 2026-04-20
fix(create-theme): support background prop in cover and intro layouts

### LINKED ISSUE #2441: Default new theme is not prepared for a background image
**Describe the bug**

When using the provided instructions to create a new theme, and adding the line 
`background: <link_to_image>`

to the `example.md` file withing the template, no image is showing up.

**Minimal reproduction**

1. Run 
    `pnpm create slidev-theme`
2. Add the line to the example.md file inside the template folder:
    `background: <link_to_image>`
4. Execute:
    `pnpm run dev`
6. The background image is not showing up

**Proposed solution**

The custom theme's cover and intro layouts are just "starters" and don't actually have the logic to display background images.

Update the files `cover.vue` and `intro.vue` to define a background prop.
Apply a dynamic style that sets the background-image if a path is provided.

Example change:

```
<script setup lang="ts">
const props = defineProps({
  background: {
    type: String,
  },
})
const style = computed(() => {
  if (props.background) {
    return {
      backgroundImage: `url(${props.background})`,
      backgroundRepeat: 'no-repeat',
      backgroundPosition: 'center',
      backgroundSize: 'cover',
    }
  }
  return {}
})
</script>
<template>
  <div class="slidev-layout cover" :style="style">
    <div class="my-auto w-full">
      <slot />
    </div>
  </div>
</template>
```

**Environment**

- Slidev version: v52.11.3
- Browser: Firefox 140.7.0esr (64-Bit)
- OS: Windows 11

hrithik18k:
Closes #2441
The default theme template's cover.vue and intro.vue layouts were missing support for the background frontmatter prop, so background images had no effect when creating a new theme.
Instead of manually computing backgroundImage inline (as suggested in the issue), this fix uses the existing handleBackground() utility from @slidev/client/layoutHelper — the same utility used by all built-in Slidev layouts like image-right and image-left. This correctly handles colors vs images, resolves local asset paths, and applies the right CSS defaults automatically.
Also added background: https://cover.sli.dev to example.md so new theme authors immediately see it working out of the box.

### DIFF
--- a/packages/create-theme/template/example.md
+++ b/packages/create-theme/template/example.md
@@ -1,5 +1,6 @@
 ---
 theme: ./
+background: https://cover.sli.dev
 ---
 
 # Slidev Theme Starter
--- a/packages/create-theme/template/layouts/cover.vue
+++ b/packages/create-theme/template/layouts/cover.vue
@@ -1,5 +1,19 @@
+<script setup lang="ts">
+import { handleBackground } from '@slidev/client/layoutHelper'
+import { computed } from 'vue'
+
+const props = defineProps({
+  background: {
+    type: String,
+    default: undefined,
+  },
+})
+
+const style = computed(() => handleBackground(props.background))
+</script>
+
 <template>
-  <div class="slidev-layout cover">
+  <div class="slidev-layout cover" :style="style">
     <div class="my-auto w-full">
       <slot />
     </div>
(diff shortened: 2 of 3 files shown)


hrithik18k:
Hey @kermanx — this fixes #2441. The default theme template's cover.vue and intro.vue were missing background prop support. Instead of the manual inline solution suggested in the issue, I used the existing handleBackground() utility from @slidev/client/layoutHelper — consistent with how all built-in layouts handle it. Also added background: https://cover.sli.dev to example.md so new theme authors see it working immediately. Ready for review!

PR #2558, opened 2026-04-24
feat: load preparser also from theme

jgosmann:
Not loading preparsers from the theme is unexpected (cp. #2402, #2456) and not documented. This PR aligns the behavior such that preparsers are also loaded from the theme. It also avoids duplicating the logic to determine the roots that things are loaded from, which might also improve perforrmance by avoiding doing `resolveAddons` twice. I hope this is in line with the [requested refactor](https://github.com/slidevjs/slidev/pull/2456#issuecomment-4223724037).

[I explained my use cases for wanting to load a preparser from a theme in a prior comment.](https://github.com/slidevjs/slidev/pull/2456#issuecomment-4222739953)

### DIFF
--- a/packages/vscode/src/projects.ts
+++ b/packages/vscode/src/projects.ts
@@ -195,7 +195,7 @@ export function removeProject(entry: string) {
 
 async function loadProject(entry: string) {
   const userRoot = dirname(entry)
-  return markRaw(await load(userRoot, entry, async (path: string) => {
+  return markRaw(await load({ userRoot, roots: [userRoot] }, entry, async (path: string) => {
     const document = workspace.textDocuments.find(d => slash(d.uri.fsPath) === path)
     if (document) {
       return document.getText()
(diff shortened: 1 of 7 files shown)


packages/vscode/src/projects.ts
```diff
@@ -195,7 +195,7 @@ export function removeProject(entry: string) {
 
 async function loadProject(entry: string) {
   const userRoot = dirname(entry)
-  return markRaw(await load(userRoot, entry, async (path: string) => {
+  return markRaw(await load({ userRoot, roots: [userRoot] }, entry, async (path: string) => {
```
jgosmann, line 198: I'm not familiar with the VSCode plugin and I'm not using it. Is it fine to just consider the `userRoot` here? `@slidev/cli` seems not accessible here to use `resolveOptions` for resolving all the roots.

But `setupPreparser` is only called in `node/cli.ts`, so I suppose the VSCode extension is not running any preparsers anyways?

kermanx (approved):
Thanks~

PR #2570, opened 2026-04-29
feat: Cloudflare Pages support for hosting

ArthurDanjou:
Auto-detect wrangler.toml or wrangler.json and generate a _headers file with long-term caching for /assets/*. 

Update _redirects output for SPA and add Cloudflare Pages docs and local Wrangler preview examples.


antfu:
Thanks for the PR! Unfortunately this doesn't work as intended: Cloudflare Pages ignores /* /index.html 200 rules (its parser rejects them as an infinite loop), and Slidev already works on Cloudflare out of the box via the 404.html fallback. The _headers part also won't trigger for most users since wrangler pages deploy doesn't require a wrangler config.

A docs-only PR noting that Cloudflare Pages works out of the box would be welcome. Thanks!

PR #2573, opened 2026-04-29
fix: restore Windows virtual style globs

### LINKED ISSUE #2567: `@slidev/conditional-styles` virtual module is empty after upgrading to 52.15.0 — theme styles & user `style.css` no longer applied
**Describe the bug**

After upgrading from `@slidev/cli@52.14.2` to `52.15.0`, the `/@slidev/conditional-styles` virtual module is generated empty:

```js
/* #__PURE__ */ Object.assign({})
/* #__PURE__ */ Object.assign({})
/* #__PURE__ */ Object.assign({})
import "/node_modules/.pnpm/katex@0.16.45/node_modules/katex/dist/katex.min.css"
```

Each of the three `Object.assign({})` corresponds to one of the `roots` (theme / addons / userRoot). The `import.meta.glob` calls produce zero matches across the board, which means:

- The user's project-root `style.css` (or `styles/index.css` / `styles.css`) is no longer loaded.
- A theme's `styles/index.css` is no longer loaded — for many themes the entire visual styling silently breaks.

Downgrading to `52.14.2` immediately restores the expected behavior.

**Suspected cause** (not fully verified)

Looking at the diff between 52.14.2 and 52.15.0, the relevant change appears to be the `makeAbsoluteImportGlob` helper, which is hinted at in the v52.15.0 release notes:

> *"Glob import should relative to self id"*

In 52.14.2:

```js
function makeAbsoluteImportGlob(userRoot, globs, options = {}) {
	const relativeGlobs = globs.map((glob) => `./${slash(relative(userRoot, glob))}`);
	const opts = { eager: true, exhaustive: true, base: "/", ...options };
	return `import.meta.glob(${JSON.stringify(relativeGlobs)}, ${JSON.stringify(opts)})`;
}
```

In 52.15.0:

```js
function makeAbsoluteImportGlob(self, globs, options = {}) {
	const relativeGlobs = globs.map((glob) => `${slash(relative(self, glob))}`);
	const opts = { eager: true, exhaustive: true, ...options };
	return `import.meta.glob(${JSON.stringify(relativeGlobs)}, ${JSON.stringify(opts)})`;
}
```

The new version is called from `templateConditionalStyles.getContent` with `self = "/@slidev/conditional-styles"` (the virtual module id). `relative(self, glob)` then produces patterns like:

```
"../../Users/Donald/Documents/.../slides/style.{ts,js,css}"
```

…and Vite's `transformGlobImport` rejects these for virtual modules:

```
Error: In virtual modules, all globs must start with '/'
```

I tried patching `makeAbsoluteImportGlob` to emit absolute-style globs (`/C:/Users/.../style.{ts,js,css}`) directly, hoping Vite would accept them — but Vite interprets a leading `/` as root-relative and joins it onto the project root again, producing nonsense paths like `<root>/C:/Users/.../style.{ts,js,css}` that match nothing. So a simple "just prepend `/`" fix does not work, and I did not investigate further.

I'm not 100% certain that the `makeAbsoluteImportGlob` change is the *only* cause — there may be additional pieces I missed — but the symptom and the diff line up cleanly with that change.

**Minimal reproduction**

My theme `slidev-theme-frankfurt` (https://github.com/MuTsunTsai/slidev-theme-frankfurt) becomes completely unstyled when upgraded to 52.15.0:

1. `git clone https://github.com/MuTsunTsai/slidev-theme-frankfurt`
2. Bump `@slidev/cli` and `@slidev/types` to `52.15.0` in `package.json`
3. `pnpm install && pnpm dev`
4. Observe that the theme's styles (defined in `styles/index.css`) are not applied to the slides.
5. Fetch `/@slidev/conditional-styles` from the dev server and observe that the resolved module body contains the empty `Object.assign({})` lines shown above instead of the expected list of style imports.
6. Downgrade to `52.14.2` and the visuals return to normal.

The path of the project does not seem to matter — I reproduced this on two different projects on Windows 11, one with a plain ASCII path (`C:\Users\XXX\Documents\Projects\Template\slidev-theme-frankfurt`) and another with spaces and non-ASCII characters in the path. Both behave the same way.

**Environment**

- Slidev version: `52.15.0` (regression) / `52.14.2` (working)
- Browser: Chrome 144 (also reproduced in Edge)
- OS: Windows 11
- Node: 22.15.1
- Package manager: pnpm
- Vite: `8.0.10`

cyphercodes:
## Summary

- Preserve the self-relative `import.meta.glob` behavior for normal paths, but fall back to Vite root-relative globs for Windows drive-letter paths so virtual modules do not lose the drive prefix.
- Pass `userRoot` into the virtual glob helper for conditional styles, global layers, and setup modules.
- Add a regression test covering `C:\\...` conditional style glob generation.

Fixes #2567.

## Tests

- `pnpm install --frozen-lockfile --ignore-scripts`
- `pnpm exec vitest run --project test test/utils.test.ts`
- `pnpm exec eslint packages/slidev/node/utils.ts packages/slidev/node/virtual/conditional-styles.ts packages/slidev/node/virtual/global-layers.ts packages/slidev/node/virtual/setups.ts test/utils.test.ts`
- `pnpm --filter @slidev/types --filter @slidev/parser --filter @slidev/client --filter @slidev/cli build`
- `pnpm exec slidev build demo/starter/slides.md --out /tmp/oss-pr-pipeline/round-10-ai-tools-20260429T231732Z/slidev-2567/validation-build-output`
- `git diff --check`

### DIFF
--- a/packages/slidev/node/utils.ts
+++ b/packages/slidev/node/utils.ts
@@ -1,7 +1,7 @@
 import type { ResolvedFontOptions, SourceSlideInfo } from '@slidev/types'
 import type MarkdownExit from 'markdown-exit'
 import type { Connect, GeneralImportGlobOptions } from 'vite'
-import { relative } from 'node:path'
+import { relative, win32 } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { slash } from '@antfu/utils'
 import { createJiti } from 'jiti'
@@ -10,6 +10,7 @@ import YAML from 'yaml'
 const RE_WHITESPACE_ONLY = /^\s*$/
 const RE_QUOTED_STRING = /^(['"])(.*)\1$/
 const RE_WHITESPACE = /\s+/g
+const RE_WINDOWS_DRIVE = /^[A-Z]:\//i
 
 type Token = ReturnType<MarkdownExit['parseInline']>[number]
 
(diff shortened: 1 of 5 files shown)


kermanx:
Thank you fpr the PR! Could you verify that `slidev build` with vite8 (with Rolldown) works correctly on Windows? The behavior of glob import resolution has somehow chakged in Rolldown.

antfu (approved):

kermanx (approved):

PR #2583, opened 2026-05-05
fix(create-slidev): avoid --open in starter dev script

### LINKED ISSUE #2582: create-slidev starter uses `slidev --open`, which can fail on Windows and make first-run setup confusing
### Describe the bug

The default starter generated by `create-slidev` uses:

```json
"dev": "slidev --open"
```

On some Windows environments, the browser auto-open step can fail with an error like:

```text
Error: spawn EPERM
```

even though the Slidev dev server itself is running correctly.

This makes the first-run experience confusing, especially for new users, because it can look like Slidev failed even when the local server is actually available.

### Steps to reproduce

1. Create a new Slidev project:
   ```bash
   npm init slidev@latest my-slides
   ```
2. Choose `npm` when prompted.
3. Run:
   ```bash
   npm run dev
   ```
4. On some Windows / restricted shell environments, browser auto-open fails with `spawn EPERM`.

### Expected behavior

The starter should provide a reliable first-run experience. Browser auto-open should not make the starter appear broken.

### Actual behavior

The dev server starts, but `slidev --open` may fail while trying to open the browser, producing an error like:

```text
Error: spawn EPERM
```

### Relevant files

- `packages/create-app/template/package.json`
- `packages/slidev/node/cli.ts`

### Possible fixes

One possible fix would be to change the starter script from:

```json
"dev": "slidev --open"
```

to:

```json
"dev": "slidev"
```

Another possibility would be to keep `--open` but make browser-open failures non-fatal and display a friendly message instead.

### Additional context

In my case, the server still started successfully and printed a valid local URL, but the browser launch failed. That made it seem like the setup was broken even though Slidev was actually running.

hrithik18k:
Closes #2582

## Summary

This updates the `create-slidev` starter template to use `slidev` instead of `slidev --open` for the generated `dev` script.

## What changed

- changed the starter template `dev` script from:
  ```json
  "dev": "slidev --open"
  ```
  to:
  ```json
  "dev": "slidev"
  ```

## Why

On some Windows or restricted shell environments, the browser auto-open step can fail with errors like:

```text
Error: spawn EPERM
```

even though the Slidev dev server itself starts successfully.

Removing `--open` from the default starter script makes the first-run experience more reliable and avoids making setup appear broken when only the browser launch failed.
```

### DIFF
--- a/packages/create-app/template/package.json
+++ b/packages/create-app/template/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "scripts": {
     "build": "slidev build",
-    "dev": "slidev --open",
+    "dev": "slidev",
     "export": "slidev export"
   },
   "dependencies": {


hrithik18k:
@antfu review please .

antfu:
I think we should fix the behaivour on Windows instead (while I don't have Windows myself, I am counting on the community to contribute). Thanks.

PR #2633, opened 2026-06-16
feat: add standalone single-file HTML bundle export

Jacobinwwey:
## Overview

This PR adds support for generating standalone single-file HTML bundles that work under `file://` protocol (double-click to open).

## Problem

Current Slidev HTML exports use ES modules with dynamic imports, which are blocked by CORS policy when opened via `file://` protocol. Users must run a local HTTP server to view presentations.

## Solution

Implement a post-build transformation that:

1. Transforms ES modules → CommonJS with custom __require loader
2. Inlines all JavaScript and CSS into single HTML file
3. Handles Vue component exports correctly for `defineAsyncComponent`
4. Transforms dynamic imports, side-effect imports, and `__vite__mapDeps` dependency resolution
5. Works offline without any server or dependencies

### Export Transformation

```
// Before (broken):
export { S as default } → module.exports.default = S

// After (fixed):
export { S as default } → module.exports.default = module.exports = S
```

### Dynamic Import & Side-Effect Import Transformation

```
// Dynamic imports:
import('./mod') → Promise.resolve(window.__require('./mod'))

// Side-effect imports:
import "./mod" → __require('./mod')
```

### __vite__mapDeps Path Resolution

Strips `assets/` prefix from dependency arrays so lazy-loaded modules (e.g., Mermaid diagram renderers) resolve correctly against the flattened module structure in the bundle.

### Hash-Based Router

Replaces Vue Router's web history mode with hash-based routing for `file://` protocol compatibility.

### CSS/Preload Handling

- Removes dynamic CSS preload functions that create link elements
- Stubs CSS preload with no-op
- Removes external preload link tags

## Verified Working

- Mermaid diagrams (all 20+ diagram types)
- Code syntax highlighting
- Slide navigation (keyboard + hash routing)
- `file://` protocol (double-click open)
- Chrome, Firefox

## Usage

```bash
slidev build --standalone-bundle
# Output: dist/index-standalone.html (double-click to view)
```

## Files Changed

- `packages/types/src/cli.ts` - Add `standalone-bundle` option type
- `packages/slidev/node/cli.ts` - Add CLI flag
- `packages/slidev/node/commands/standalone-bundler.ts` (new) - Core bundler
- `packages/slidev/node/commands/build.ts` - Integration
- `test/standalone-bundler/standalone-bundler.test.ts` (new) - Test suite

## Breaking Changes

None - opt-in feature via `--standalone-bundle` flag.

### DIFF
--- a/PR_DESCRIPTION.md
+++ b/PR_DESCRIPTION.md
@@ -1,0 +1,136 @@
+# Standalone Bundle Export for Slidev
+
+## Overview
+
+This PR adds support for generating standalone single-file HTML bundles that work under `file://` protocol (double-click to open).
+
+## Problem
+
+Current Slidev HTML exports use ES modules with dynamic imports, which are blocked by CORS policy when opened via `file://` protocol. Users must run a local HTTP server to view presentations.
+
+## Solution
+
+Implement a post-build transformation that:
+
+1. Transforms ES modules → CommonJS with custom loader
+2. Inlines all JavaScript and CSS into single HTML file
+3. Handles Vue component exports correctly for `defineAsyncComponent`
+4. Works offline without any server or dependencies
+
+## Technical Details
+
+### Key Fix: Export Transformation
+
+The critical fix ensures Vue components load correctly:
+
+```javascript
+// Before (broken):
+export { S as default };
+  ↓
+module.exports.default = S;  // Only .default set
+
+// After (fixed):
+export { S as default };
+  ↓
+module.exports.default = module.exports = S;  // Dual assignment
+```
+
+**Why this matters:** Vue's `defineAsyncComponent` checks both `component.default` and `component` (as fallback). CommonJS `require()` returns the entire `module.exports` object, so both must point to the component for proper compatibility.
+
+### Module System
+
+Custom `__require()` function provides:
+
+- Module caching
+- Relative path resolution
+- CommonJS semantics
+- Error handling
+
+### Browser Compatibility
+
+- ✅ Chrome/Chromium
+- ✅ Firefox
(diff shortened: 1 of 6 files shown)


Jacobinwwey:
### Recent fixes (2026-06-17)

Three additional fixes pushed to `feat/standalone-bundle-export`:

**1. Transform side-effect imports**
`import "./mod"` and `import './mod'` now transformed to `__require('./mod')`. Previously these bare imports were left untransformed, causing "Cannot use import statement outside a module" errors when the code was eval'd via `new Function()`.

**2. Strip `assets/` prefix from `__vite__mapDeps` dependency arrays**
Mermaid and other lazily-loaded modules use local `__vite__mapDeps` functions with hardcoded paths like `m.f=["assets/c4Diagram-xxx.js"]`. The bundler now strips the `assets/` prefix from all string literals in module code since modules are flattened to root level in the bundle.

**3. Global `__vite__mapDeps` stub with path normalization**
Replaced the `return []` stub with a proper implementation that maps indices to paths and also strips `assets/` prefix for cases where the global stub (not local) is used.

All fixes verified working with Playwright automated testing across all slides including Mermaid diagrams.

antfu:
Thanks for the PR. But I think we can't take such complexity. The limitation of the module script is from Vite - this kind of patching can be unstable and hard to maintain. I would suggest you to create a general Vite plugin to do the postprocess, and then you might share the plugin with the community, so every Vite users might be benefit from it, and you can also apply the plugin to Slidev easily. That said, I would generally recommend using a simple SPA hosting script to do so instead of hacking the HTML and scripts.

PR #2640, opened 2026-06-19
feat(skill): add slide reorder/remove script for agentic use

nor3th:
Hi

I have found the approach quite great to develop the slides through agentic AI interaction using claude code. However for some tasks (swap slides, move slide 10 after the current slide 20 or delete slide 10) claude code used too many tokens and took too long, hence I let it develop those skill additions which make those tasks cheaper and faster. 

I am not sure if it is in the project's scope to improve the claude code skills as well or a claude code skill and any other future additions should be conducted in separate repositories. Just let me know if additions like those are useful or if I should just keep them to my own fork.

I haven't written any test cases yet, but I can still work on those if necessary.

**Summary**

Adds scripts/slide_ops.py to the Slidev skill with three operations:
- swap N M     — swap two slides by 1-based index
- move N after|before M — reorder a slide relative to another
- remove N     — delete a slide

Manually reordering slides via the Edit tool requires reading ~200 lines verbatim and constructing a large replacement — expensive in tokens and error-prone with Slidev's compound separators (---\nlayout: x\n---). The script handles both simple and compound separators correctly and runs in milliseconds.

Documents usage in SKILL.md so Claude picks it up automatically and responds to natural language like "swap slides 3 and 7".

### DIFF
--- a/skills/slidev/README.md
+++ b/skills/slidev/README.md
@@ -21,6 +21,7 @@ The Slidev skill provides Claude Code with knowledge about:
 - **Layouts** - Built-in layouts, slots, global layers
 - **Presenter Mode** - Recording, timer, remote access
 - **Exporting** - PDF, PPTX, PNG, SPA hosting
+- **Slide Structure Operations** - Swap, move, and delete slides by index via a bundled script
 
 ## Usage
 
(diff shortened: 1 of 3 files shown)


antfu:
Thank you, I do like this idea but I would prefer to use our existing parser and the same js ecosystem, and then we might be able to do expose some tools via our cli instead of python script. I think we already have some sort of MCP support for the VS Code extension; bringing them to the CLI and the Web server might be a good idea (cc @kermanx).

I am closing this PR for now but I am open for contributions around this idea.

PR #2651, opened 2026-06-25
feat: offline support via opt-in PWA precaching

quinnypig:
## Summary

Adds an opt-in `offline` headmatter flag. When enabled, `slidev build` generates a service worker (via `vite-plugin-pwa` / Workbox) that **precaches every deck asset — JS/CSS/HTML plus all images, video, and audio** — so a built, *served* deck works fully offline after the first load.

- `offline: false` (default) — no service worker
- `offline: true` — enabled in dev and build
- `offline: 'build'` — built output only

A small bottom-right indicator shows "Caching for offline…" then briefly "Ready offline" when precaching completes. The build plugin is a no-op when disabled, and the client registration + indicator tree-shake out via a `__SLIDEV_FEATURE_OFFLINE__` build flag (mirrors the existing `wakeLock` flag pattern).

## Why

For presenting on an unknown/locked-down machine over a flaky or absent network: serve the deck once, let it cache, then it runs entirely from the service worker — no per-slide asset fetching mid-talk.

## Verified

Built a deck with `offline: true` (an image **and** a 2.9 MB video), served it over localhost, then **disabled the network and reloaded** — the deck, image, and video all load from the service-worker cache (44 precached entries including the video). Workbox `maximumFileSizeToCacheInBytes` is raised to 100 MB and the glob covers image/video/audio extensions so large media isn't silently skipped.

## Notes / follow-ups

- **Video seeking offline:** precached media is served as a full cached response; play-from-start works offline (verified), but scrubbing/seeking may want a Workbox Range-requests handler — happy to add if desired.
- **Remote assets:** only files emitted into the build are precached; CDN/remote-fetched assets won't be offline. The manifest intentionally ships no icons (an offline feature shouldn't depend on a remote icon).
- Default off; opt-in and heavy. Relates to discussion #2644 (where offline support should live — core vs addon); happy to reshape as an addon if preferred.

## Verification

`pnpm build` + `pnpm test` + `pnpm typecheck` + `pnpm lint` all pass.

### DIFF
--- a/packages/types/src/frontmatter.ts
+++ b/packages/types/src/frontmatter.ts
@@ -227,6 +227,12 @@ export interface HeadmatterConfig extends TransitionOptions {
    */
   wakeLock?: boolean | 'dev' | 'build'
   /**
+   * Enable offline support: a service worker precaches all deck assets on first
+   * load so a served deck works fully offline afterward. Heavy; opt-in.
+   * @default false
+   */
+  offline?: boolean | 'dev' | 'build'
+  /**
    * Force the filename used when exporting the presentation.
    * The extension, e.g. .pdf, gets automatically added.
    *
--- a/packages/types/src/frontmatter.ts
+++ b/packages/types/src/frontmatter.ts
@@ -227,11 +227,11 @@ export interface HeadmatterConfig extends TransitionOptions {
    */
   wakeLock?: boolean | 'dev' | 'build'
   /**
-   * Enable offline support: a service worker precaches all deck assets on first
+   * Enable PWA support: a service worker precaches all deck assets on first
    * load so a served deck works fully offline afterward. Heavy; opt-in.
    * @default false
    */
-  offline?: boolean | 'dev' | 'build'
+  pwa?: boolean | 'dev' | 'build'
   /**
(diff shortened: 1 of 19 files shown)


packages/types/src/frontmatter.ts
```diff
@@ -226,6 +226,12 @@ export interface HeadmatterConfig extends TransitionOptions {
    * Enable wake lock
    */
   wakeLock?: boolean | 'dev' | 'build'
+  /**
+   * Enable offline support: a service worker precaches all deck assets on first
+   * load so a served deck works fully offline afterward. Heavy; opt-in.
+   * @default false
+   */
+  offline?: boolean | 'dev' | 'build'
```
antfu: Maybe it would be clearer to call this feature "pwa"?

quinnypig: Good call — it *is* a PWA, and that's the clearer name. Renamed the `offline` headmatter option → `pwa` (plus the `__SLIDEV_FEATURE_PWA__` build flag, the indicator component, and the regenerated schema) in d7d5d719. Kept the user-facing "Caching for offline…" / "Ready offline" wording since that's the benefit people actually recognize.

Issue #2629, opened 2026-06-12
Router path duplication with --base in Slidev v52


mrsimpson:
## Bug Description

When using --base /path/ with Slidev v52.16.0, navigating to the next slide causes the URL to incorrectly duplicate the base path: /path/path/2 instead of /path/2.

## Minimal Reproduction

git clone https://gist.github.com/mrsimpson/d44bb8c60d9540450f99d88a166eab17.git
cd slidev-bug-repro
npm install
npm run build -- --base /test-slides/
npx vite preview --base /test-slides/

Open http://localhost:4173/test-slides/1 and click 'next slide'.

Expected: http://localhost:4173/test-slides/2
Actual: http://localhost:4173/test-slides/test-slides/2

## Environment

- Slidev version: 52.16.0
- Node.js: 20/24
- OS: macOS/Windows/Linux

## Notes

- This bug only occurs with npm run build + vite preview, not with npm run dev
- Works correctly in Slidev 0.50.x
- Affects GitLab Pages deployments with subdirectory paths

inaku-Gyan:
I met the same issue, and tried to fix it in #2630.

ifigueroap:
Just upgraded to 0.52 and got hit by this bug too. Claude is making a [patch-package](https://www.npmjs.com/package/patch-package) fix as we speak:

```
diff --git a/node_modules/@slidev/client/logic/slides.ts b/node_modules/@slidev/client/logic/slides.ts
index cb156a1..482111a 100644
--- a/node_modules/@slidev/client/logic/slides.ts
+++ b/node_modules/@slidev/client/logic/slides.ts
@@ -21,7 +21,7 @@ export function getSlidePath(
     route = getSlide(route)!
   const no = route.meta.slide?.frontmatter.routeAlias ?? route.no
   const path = exporting ? `export/${no}` : presenter ? `presenter/${no}` : `${no}`
-  return `${import.meta.env.BASE_URL}${path}`
+  return `/${path}`
 }
 
 export function useIsSlideActive() {
```

barabasss:
Confirmed on a real GitHub Pages project-site deploy (`--base /<repo>/`, history routing, Slidev 52.16.0). Paging to the next slide produces `/<base>/<base>/2` and Slidev's NotFound; direct deep-links work, and `dev` works (base `/`).

[...]

PR #2655, opened 2026-07-01
feat: add Typst as an alternative formula renderer

### LINKED ISSUE #2157: Add Typst support as another option for formula rendering
Hi, Slidev is really nice as an alternative to not only the PowerPoint but also $\LaTeX$ beamer

[Typst](https://github.com/typst/typst) is another brand-new scientific typesetting system as a replacement to $\LaTeX$, with powerful and modern math formula typesetting support but with more concise syntax. and a better package manager support which enables one to use third party(to Typst) package to do more complicated task, like plotting, drawing vector graphics, etc. 

Currently KaTeX is the only math formula renderer for Slidev. Typst could be another option for the same purpose. Especially for those who are documenting with Typst, it will bring a consistent experience. 

[Typst.ts](https://myriad-dreamin.github.io/typst.ts/cookery/guide/all-in-one-node.html) is a project introducing the typst to the JS world, which makes it easy to rendering typst source code to SVG or HTML(formula will be in MathML, experimental support currently) in both server side and client side.

 I would be appreciative if someone consider integrating this to Slidev! 😊

GZGavinZhao:
Fixes #2157.

Allows writing Typst in both inline math expressions `$ $` and multi-line math blocks `$$ $$`.

Previously implemented in #2183, but that was done when Typst doesn't have native HTML support. The recently release Typst v0.15 supports emitting MathML for HTML-native math expressions. The MathML are emitted at build-time, so this adds no runtime dependencies.

The build-time dependency is kind of huge though (a ~50M native Typst compiler executable is needed). Therefore, like the playwright feature, it's added to `peerDependencies`. Slidev users who want to use Typst need to manually add it as a dependency via `npm i -D @myriaddreamin/typst-ts-node-compiler`.
```json
  "peerDependencies": {
    "@myriaddreamin/typst-ts-node-compiler": "0.8.0-rc2",
    "playwright-chromium": "^1.10.0"
  },
  "peerDependenciesMeta": {
    "@myriaddreamin/typst-ts-node-compiler": {
      "optional": true
    },
    "playwright-chromium": {
      "optional": true
    }
  },
```

[Demo slides](https://gzgavinzhao.github.io/slidev-typst-demo/) with [source](https://github.com/GZGavinZhao/slidev-typst-demo/blob/main/slides.md).

~~Commit history contains a revert of a buggy commit (due to #2629, waiting on #2630) purely for personal testing; will revert before requesting a review.~~ Fix has been merged; the commit history is now clean.


GZGavinZhao:
Just to note: this is not stale and is actually ready for use! I'm just doing more stress testing as I'm dogfooding this PR by using Typst in a pretty important presentation, noting any weird bugs or inconveniences along the way.

peekxc:
Maybe related (or not): 
- [typst-wasm](https://www.npmjs.com/package/typst-wasm)
- [vite-plugin-typst](https://www.npmjs.com/package/@typst-wasm/vite-plugin-typst)

GZGavinZhao:
Thanks @peekxc! The WASM approach is interesting. My rationale for using a native Typst binary from [`typst-ts-node-compiler`](https://www.npmjs.com/package/@myriaddreamin/typst-ts-node-compiler) is that, given all of the Typst -> MathML processing happens on server side / build time, the easiest and most straightforward way to implement this feature is to just call the Typst compiler. I didn't spend effort into figuring out a truly cross-platform solution like WASM since that no matter platform you use, Typst is almost guaranteed to have a binary for that platform.

I'm not opposed to either direction, so I'm open to what the reviewers think. I might be wrong but I'm a little inclined to think that typst-wasm is less battle tested than Myriad-Dreamin's typst.ts, maybe that's part of the equation (pun not intended)?

Anyways the presentation went really well with the Typst slides. The only troublesome part that I found:
- The default font that Typst uses (STIX Two Math) is not supported well through Google Fonts, the web font provider that Slidev uses (https://github.com/google/fonts/issues/7953#issuecomment-4884869050). We may need to bundle the font, and it seems like `typst-wasm/fonts` package has already done some work for us? Will look into this deeper.
- I thought I inserted a placeholder div/element to replace a Typst expression when the compilation for that expression fails so it doesn't halt the build, but I don't always see that div. Given that I shouldn't be writing incorrect Typst formulas in the first place I didn't solve this when making my presentation. Now that I'm making this merge-ready it's time to fix this issue.

Will try to fix those over the weekend, then this should be ready to review!

antfu:
Would looking forward to it!

PR #2684, opened 2026-07-26
chore: Supply Chain Attack Countermeasures in GitHub Actions

ysknsid25:
Using tags like `@v6` in GitHub Actions defaults to the latest version, which carries the risk of inadvertently using a version with vulnerabilities; pinning the action to a specific SHA helps mitigate supply-chain attack risks. Additionally, using `actions/checkout@v7` offers [further benefits](https://github.blog/changelog/2026-06-18-safer-pull_request_target-defaults-for-github-actions-checkout/).

### DIFF
--- a/.github/workflows/autofix.yml
+++ b/.github/workflows/autofix.yml
@@ -16,13 +16,13 @@ jobs:
     timeout-minutes: 10
 
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v4
+        uses: pnpm/action-setup@0ebf47130e4866e96fce0953f49152a61190b271 # v6.0.9
 
       - name: Use Node.js lts/*
-        uses: actions/setup-node@v6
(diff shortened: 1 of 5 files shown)


antfu:
Can you introduce something like https://github.com/sxzz/actionspack so we can maintain the update easier?

ysknsid25:
@antfu 
The safest approach is likely to use Dependabot to handle GitHub Actions updates. It eliminates the need for commit tokens and reduces the potential supply chain attack vectors associated with using third-party libraries.

With Dependabot, for instance, the following configuration is all that's required.

```yml
version: 2
updates:
  - package-ecosystem: github-actions
    directory: /
    schedule:
      interval: weekly
    cooldown:
      default-days: 7
    commit-message:
      prefix: "dependabot"
    groups:
      github-actions:
        update-types:
          - "patch"
          - "minor"
        patterns:
          - '*'
```

What do you think?

antfu:
I just implemented the update check for GitHub Actions in taze: https://github.com/antfu-collective/taze/pull/304, where they can be bumped just like other deps. The reason I don't like Renovate or Dependabot is that they create too much noise, especially in my case where I work on a huge number of projects; going through each update PR is fundamentally impossible. I think we can have this PR merged as-is as taze would also check sha pinned actions. Thanks!

PR #2703, opened 2026-08-13
fix(create-app): default manual setup to npm

NgoQuocViet2001:
## Summary

When `create-slidev` is run directly with Node, no package manager may be detected. Choosing to install later then produced `null install` and `null run dev` in both the generated README and terminal output.

- replace the custom detection logic with `package-manager-detector`
- fall back to npm only for the manual setup path
- keep the generated README and printed commands aligned

## Tests

- `pnpm --filter create-slidev build`
- `pnpm exec eslint packages/create-app/index.mjs packages/create-app/package.json`
- `pnpm test`
- manually ran `create-slidev` without a package-manager user agent and verified the terminal and generated README use npm


antfu:
Let's use https://github.com/antfu-collective/package-manager-detector and fallback to npm. Would you like to update the PR? Thanks

NgoQuocViet2001:
@antfu Updated — package manager detection now uses `package-manager-detector`, with npm kept as the fallback for manual setup. Thanks!

packages/create-app/utils.mjs
```diff
@@ -0,0 +1,16 @@
+/**
+ * @param {string | null} packageManager
+ */
+export function getManualPackageManager(packageManager) {
+  return packageManager ?? 'npm'
```
antfu: Please minimize the code changes. I don't think this file is necessary

packages/create-app/index.mjs
```diff
@@ -95,21 +96,7 @@ async function init() {
...
-    const userAgent = process.env.npm_config_user_agent || ''
-    const execPath = process.env.npm_execpath || ''
-    if (execPath.includes('pnpm') || userAgent.includes('pnpm'))
-      pm.push('pnpm')
-    if (execPath.includes('yarn') || userAgent.includes('yarn'))
-      pm.push('yarn')
-    return pm.length === 1 ? pm[0] : null
-  }
-  const pkgManager = getPkgManager()
+  const pkgManager = getUserAgent()
```
antfu:
```suggestion
  const pkgManager = getUserAgent() ?? 'npm'
```

NgoQuocViet2001: @antfu Updated — the npm fallback now applies directly to the detected package manager.

PR #2714, opened 2026-08-21
fix: apply dev define flags to JS/TS client modules

### LINKED ISSUE #2673: Unable to add new directives in `setup/main.ts`
**Describe the bug**

I've added a copy of `v-mark` with different name and added it in `main.ts` and `app.use`.
Cold start shows an error: `[Vue warn]: Failed to resolve directive: instantmark`.
If I resave the file, the component starts to work without errors.

**Minimal reproduction**

1. create `setup/main.ts`
2. Copy definition of `v-mark` and change the name for the directive.
3. Add plugin using `app.use`
4. Start slidev and see error in browser console.
5. Resave `main.ts`

**Environment**

- Slidev version: v52.16.0
- Browser: Chrome
- OS: MacOS

SomSamantray:
## Summary

Custom directives registered via `setup/main.ts` (for example a `v-mark` copy renamed to `v-instantmark`) failed to resolve on a cold dev-server start with `[Vue warn]: Failed to resolve directive`, and only started working after resaving the setup file.

The cause is in Vite 8: its `vite:define` does not substitute bare `define` identifiers for the client environment in dev (see the vitejs/vite#22419 class of regression), so `__DEV__` and `__SLIDEV_*` flags passed through unsubstituted in `.ts`/`.js` client modules such as `packages/client/setup/main.ts`. That threw at router creation before the custom setup ran, so the directive never registered.

The fix extends the existing `slidev:flags` transform plugin (previously Vue-SFC-only) to also cover `.ts`/`.js` modules in dev:

- Identifier-boundary replacement so compound identifiers and string content are not corrupted.
- JS/TS substitution is gated to dev; Vue SFC handling stays active in every mode because compiled render functions reference flags as property accesses that Vite's identifier-only define cannot substitute in production builds either.
- The id filter is narrowed to `.vue`, `?vue&type=script`, and `.ts`/`.js`, excluding `node_modules`.
- The static pre-check and single regex pass are derived from the actual define keys.

Fixes #2673

## Validation

- New unit tests cover TS/JS modules, Vue SFC (dev and build), query-suffixed ids, `node_modules` exclusion, build-mode gating, and identifier-boundary behavior: 11 tests pass.
- Full `@slidev/cli` suite: 72 tests pass; `pnpm lint` and `pnpm typecheck` clean.
- Browser-verified with a real dev-server cold start: a deck with a custom `setup/main.ts` directive renders and the directive's annotation appears with no console errors; production build output contains no raw `__SLIDEV_*` identifiers.

### DIFF
--- a/packages/slidev/node/vite/compilerFlagsVue.test.ts
+++ b/packages/slidev/node/vite/compilerFlagsVue.test.ts
@@ -1,0 +1,80 @@
+import type { ResolvedSlidevOptions } from '@slidev/types'
+import { describe, expect, it } from 'vitest'
+import { createVueCompilerFlagsPlugin } from './compilerFlagsVue'
+
+function makeOptions(define: Record<string, string>): ResolvedSlidevOptions {
+  return {
+    utils: {
+      define,
+    },
+  } as unknown as ResolvedSlidevOptions
+}
+
+function transformCode(define: Record<string, string>, code: string, id: string) {
+  const plugin = createVueCompilerFlagsPlugin(makeOptions(define))
+  const handler = (plugin.transform as { handler: (code: string, id: string) => string | undefined }).handler
+  return handler(code, id)
+}
+
+const DEFINE = {
+  __DEV__: 'true',
+  __SLIDEV_HASH_ROUTE__: 'false',
+  __SLIDEV_MEMORY_ROUTE__: 'false',
+  __SLIDEV_FEATURE_PWA__: 'false',
+}
+
+describe('createVueCompilerFlagsPlugin', () => {
+  it('replaces define flags in TypeScript modules', () => {
+    const code = `
+const router = createRouter({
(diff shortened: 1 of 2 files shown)


SomSamantray:
The failing check here (`test (ubuntu-latest, pnpm, true)` in **Production Smoke Test**) is unrelated to this PR — it's a pre-existing break on `main`:

```
[plugin unplugin-vue-markdown] slides.md__slidev_2.md
Error: Renderer.renderInline: async rule detected, use renderInlineAsync()
```

The last 4 runs of the Production Smoke Test workflow on `main` (starting ~2026-08-24T04:01Z) all fail the same way, with the last green run back on 2026-08-19. Looks like a `markdown-exit`/`@comark/markdown-it` version mismatch causing a sync `renderInline` call where an async rule now requires `renderInlineAsync`. Not touching anything this PR changes (`packages/client/setup/main.ts`, the `slidev:flags` plugin) — flagging so it's not mistaken for a regression from this change.

Issue #2721, opened 2026-08-26
feat(export): editable PPTX export with native shapes and text


stn1slv:
### The problem

`slidev export --format pptx` renders each slide to a PNG and assigns it as `slide.background` ([`export.ts:530-532`](https://github.com/slidevjs/slidev/blob/main/packages/slidev/node/commands/export.ts#L530-L532)). Every `ppt/slides/slideN.xml` in the result is 770 bytes with an empty `<p:spTree>`, so the deck is a stack of pictures. Nothing can be selected, retyped or recolored. `guide/exporting.md` states this as a known limitation.

That is the right default for visual fidelity, and I am not proposing to change it. But it means the answer to "can you send me the deck so I can adjust a few slides?" is currently no. Two open reports ask for it: [discussion 2417](https://github.com/slidevjs/slidev/discussions/2417) and #2637.

Right-clicking the text of a slide today offers New Slide, Duplicate Slide and Format Background, with Cut and Copy greyed out. There is no object under the cursor, because the words are part of the slide's background.

<img width="2412" height="1354" alt="Image" src="https://github.com/user-attachments/assets/7b653c07-c02b-4f13-acc7-5f51615262a3" />

_(screenshot: the right-click menu on a slide exported today)_

The same right-click after this change offers Font, Paragraph, Bullets and Hyperlink, on a selected text box with a caret in the line.

<img width="2400" height="1279" alt="Image" src="https://github.com/user-attachments/assets/3a46bd83-b32f-4106-b32a-cc57692e3dce" />

_(screenshot: the same right-click with this change)_

<img width="1197" height="669" alt="Image" src="https://github.com/user-attachments/assets/6c376498-648f-42f0-aab1-64960e0f8704" />

_(screenshot: select all (crtl+a) for the slide)_

### What I would like to contribute

A fifth export format, `--format pptx-editable`, which walks the rendered DOM at export time and emits native PowerPoint shapes and text boxes instead of one picture per slide.

On `demo/starter`, unmodified:

| | `--format pptx` | `--format pptx-editable` |
|---|---|---|
| slides | 40 | 40 |
| `slide1.xml` | 770 bytes | 5,134 bytes |
| shapes | 0 | 640 |
| text runs | 0 | 1,554 |
| pictures | 0 | 66 |

Side by side the two files render almost identically, so nothing is given up to get there:

[...]

PR #2722, opened 2026-08-26
feat(export): editable PPTX export with native shapes and text

stn1slv:
Closes the request in #2721, which has the full reasoning. This is the implementation.

## Try it before reading any code

```bash
# once CI has run on this PR and pkg.pr.new has published
npm i https://pkg.pr.new/@slidev/cli@2722 playwright-chromium
slidev export --format pptx-editable
```

Then click on a heading in the result. It is text.

The workflows on this PR are held at `action_required`, so nothing has been published yet. Approving them is what produces that build.

If you would rather not approve workflows first, the same thing from a checkout, no CI involved:

```bash
gh pr checkout 2722
pnpm install && pnpm build
node packages/slidev/bin/slidev.mjs export demo/starter/slides.md --format pptx-editable
```

## What this does not do

- No template or `.potx` loading, no slide master synthesis, no font embedding.
- A `.pptx` names fonts rather than carrying them, and the name written is the resolved family with any `@fontsource` `Variable` suffix removed, so a theme built on `Inter Variable` asks for `Inter`. A recipient is likelier to have the static face installed, and naming the variable family would only move the substitution to their machine. The export prints which families it referenced.
- No gradients: `pptxgenjs` has none, so they rasterize.
- `--per-slide` errors for this format. It navigates one slide at a time and never renders the print page the measurement depends on.
- Decorations a theme draws with `::before` or `::after` in normal flow are left out, and named at the end of the export. Code block line numbers are the case a user is most likely to meet: they come from a CSS counter, which has neither text nor a box a computed style can report.
- PowerPoint draws a dashed underline as a solid one, so a theme that rules its links with `border-bottom: 1px dashed`, as Slidev's own do, gets a solid rule. The style is in the file as `u="dash"` and other renderers honor it.
- A long paragraph can wrap onto a different number of lines, because PowerPoint's text metrics are not Chromium's.

## What it does

`--format pptx-editable` walks the rendered DOM at export time and emits native shapes and text boxes rather than one picture per slide. On `demo/starter`, unmodified:

| | `--format pptx` | `--format pptx-editable` |
|---|---|---|
| `slide1.xml` | 770 bytes | 5,134 bytes |
| shapes | 0 | 640 |
| text runs | 0 | 1,554 |

`--format pptx` is untouched. The two ship side by side because `pptx` is pixel-exact on every theme and this cannot be.

## Shape of the change

`export.ts` changes by twenty-eight lines: one entry in the format union, one dispatch branch, the format cast, `withClicks ?? format === 'pptx'` becoming `format.startsWith('pptx')`, and a generator that guards `--per-slide`, delegates, and hands the result to a reporter. One line in `cli.ts` for `choices`. The rest is new files under `packages/slidev/node/commands/pptx/`:

- `walker.ts` runs in the page and extracts only facts that need live layout. It makes no decisions.
- `normalize.ts` holds every judgement as pure functions over plain JSON.
- `color.ts` parses computed colors in every syntax Chromium leaves in a computed style.
- `capture.ts` is the only Playwright glue.
- `build.ts` turns the result into a `.pptx` through `pptxgenjs`, already a `catalog:prod` dependency. Also pure.

115 tests, over hand-written fixtures and unzipped OOXML rather than over the calling code. `plans/022` asks for the first tests the export pipeline has ever had, is P1 and still TODO, and `plans/023` is blocked on it. The split here is the shape `023` wants, so the other exporters can be extracted later without touching this one.

One dev dependency, `jszip`, added through the catalog. It is used only to unzip and assert on the generated XML in tests.

## The safety valve

A slide that throws, times out, loses all its text, or comes out more than 60% picture with nothing recovered over it falls back to exactly today's `--format pptx` output, for that slide alone, and says why. None of the decks I have run needs it. It is there so that a theme this heuristic cannot handle degrades to current behavior, per slide and loudly, instead of producing a broken file.

## On whether it survives real themes

It did not, at first. I built it against `demo/starter`, then ran it over [`peacock0803sz/slides`](https://github.com/peacock0803sz/slides), [`lyqht/intro-to-svg-slides`](https://github.com/lyqht/intro-to-svg-slides) and two decks on custom themes of my own, written to match internal corporate templates, which I cannot share. Four distinct themes in all, none of which I wrote the exporter against. Every slide was compared to its rendered reference. Twenty-six distinct defects came out of that, each now a named test: paint order is not array order, an inline background is painted once per line fragment rather than once over their union, `locator.screenshot()` returns the wrong region on a page far taller than its viewport, KaTeX cannot be walked as text, a table cell is not a layout container, and so on.

So when a theme breaks it, `normalize.ts` is pure and the report becomes a fixture plus a test. Nobody has to reproduce it inside a browser first.

## Questions I would rather settle in review

1. Is `--format pptx-editable` the surface you want, or would you rather this arrived as a pluggable export-format API with this as its first consumer? Everything already sits behind a `PptxExportContext` and is a plugin in all but name.
2. Is the per-slide image fallback the right degradation policy?

I am happy either way, and would rather rework the surface now than after review.

### DIFF
--- a/docs/builtin/cli.md
+++ b/docs/builtin/cli.md
@@ -70,7 +70,7 @@ Export slides to PDF (or other format). See <LinkInline link="guide/exporting" /
 Options:
 
 - `--output` (`string`, default: use `exportFilename` (see https://sli.dev/custom/#frontmatter-configures) or use `[entry]-export`): path to the output.
-- `--format` (`'pdf', 'png', 'pptx', 'md'`, default: `'pdf'`): output format.
+- `--format` (`'pdf', 'png', 'pptx', 'pptx-editable', 'md'`, default: `'pdf'`): output format. `pptx` exports each slide as an image; `pptx-editable` rebuilds them as native PowerPoint shapes with selectable text.
 - `--timeout` (`number`, default: `30000`): timeout for rendering the print page (see https://playwright.dev/docs/api/class-page#page-goto).
 - `--range` (`string`): page ranges to export (example: `'1,4-5,6'`).
 - `--dark` (`boolean`, default: `false`): export as dark theme.
--- a/docs/guide/exporting.md
+++ b/docs/guide/exporting.md
@@ -90,6 +90,7 @@ Two things worth knowing before you send the file on:
 
 - A `.pptx` names fonts, it does not embed them. The export prints which font families it referenced; recipients need those installed or PowerPoint will substitute.
 - PowerPoint does not measure text exactly as a browser does, so a long paragraph may wrap onto a different number of lines.
+- Decorations a theme draws with `::before` or `::after` in normal flow are left out, and the export lists them. Code block line numbers are one of these: they come from a CSS counter, which has no text and no box that a computed style can report.
 
 `--per-slide` is not supported with this format.
 
--- a/docs/guide/exporting.md
+++ b/docs/guide/exporting.md
@@ -72,6 +72,27 @@ Note that all the slides in the PPTX file will be exported as images, so the tex
 
 In this mode, the `--with-clicks` option is enabled by default. To disable it, pass `--with-clicks false`.
 
+### Editable PPTX
+
+If the recipient needs to edit the deck rather than only present it, export it with native shapes instead of pictures:
+
+```bash
+$ slidev export --format pptx-editable
+```
+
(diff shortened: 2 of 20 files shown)


antfu:
This is amazing! Thank you!

PR #2739, opened 2026-09-09
fix(cli): handle publicIpv4() errors when resolving public IP

### LINKED ISSUE #2724: Add error handling for calls to publicIpv4 to prevent crashing on --remote=<password> if public IP cannot be determined.
**Describe the bug**

Error handling (try/catch) is missing for calls to publicIpv4(), meaning that slidev crashes (DOMException) if the public IP cannot be determined. 
This happens in practice when using slidev --remote with a password present, or when asking for a QR code, on a server that has a LAN IP but which cannot determine a public IP due to e.g. network restrictions on DNS.

**Minimal reproduction**
- Any presentation
- Block DNS calls (UDP/53 to at least opendns + google, which seems to be used by public-ip)
- bun dev --remote=abcd1234

**Environment**

- Slidev version: 52.19.0
- Browser: NA
- OS: Ubuntu 26.04.1 LTS

jabrailkhalil:
## Summary

`publicIpv4()` from the `public-ip` package rejects when the public IP cannot be determined (e.g. DNS restricted networks, no internet). Two call sites in `packages/slidev/node/cli.ts` awaited it without error handling:

- startup with `--remote=<password>` (crashed the whole CLI)
- the `c` QR code shortcut (crashed the shortcut handler)

Both calls are now wrapped in try/catch. When the lookup fails, the CLI keeps running: the LAN addresses printed by `printInfo()` remain available, and the QR shortcut simply skips the "Public IP" line.

## Tests

- `pnpm typecheck` - passed (vue-tsc, exit 0)
- `pnpm build` - passed (all packages)
- `pnpm test` - 241/241 passed (28 test files)
- `pnpm exec eslint packages/slidev/node/cli.ts` - no issues

## Verification

Manual scenario from the issue: launch a deck with `slidev --remote=abcd1234` on a machine where the public IP cannot be determined. Before this change the CLI throws an unhandled `DOMException` and dies; after the change it prints the local remoting URLs and stays interactive. The QR code shortcut (`c`) no longer throws either and only skips the public IP line.

Fixes #2724


jabrailkhalil:
The CI workflows for the current head `57b84289ee` are waiting for contributor workflow approval (`action_required`): [Production Smoke Test](https://github.com/slidevjs/slidev/actions/runs/34404587882), [Test](https://github.com/slidevjs/slidev/actions/runs/34404587883), [autofix.ci](https://github.com/slidevjs/slidev/actions/runs/34404587912). Could a maintainer approve the pending runs when convenient? These runs have not executed their checks yet; I will address any failures once they run.

